PR #17, opened 2016-03-13
Schema specifies use String not utf8

stheppi:
Allowing the user to specify if the schema generated should specify use String not Utf8. This way when Avro deserialises the payload will create String instances not Utf8

Not sure what the best way is to change the macro ToSchema to allow a flag. I leave that to your capable hands :)


sksamuel:
Nice. Should maybe add similar options for avro reader.

Issue #21, opened 2016-05-11
Compatibility issue Serializing/Deserializing messages generated with avro4s and other languages


annovi:
I have two separate applications that  interact using a queue in the middle, and they communicate sending messages to each other, using Avro (Serialize and Deserialize).
The queue in the middle is managed by a message broker (RabbitMq)

The scenario is really simple: 
- one application create a message (based of a schema), serialize the message and put the message in the queue.
- on the other side there is an other program that pick this message, deserialize the avro message and read the content.

This is my schema (really simple)
```
{
	"type": "record",
	"name": "Surface",
	"fields": [
	{
		"name": "surfaceId",
		"type": "string"
	}]
}
```
`case class Surface(surfaceId: String)`

I serialize the message using avro4s library and I send this message to  message broker.
I use this code to serialized the message:

**First scenario**: two applications made in **Scala**
```
    val surface = Surface(surfaceId = "5423g343423")
    val stream = new ByteArrayOutputStream()
    val os = AvroOutputStream[Surface](stream, true)
    os.write(surface)
    os.flush()
```
On the other side there is an other application that grab the message from the queue, and deserialize the message with the avro4s library. Body is Array[Byte]

```
        val is = AvroInputStream[Surface](body)
        val result = is.iterator.toSet
        println(result.mkString("\n"))
        is.close()
```

This configuration works perfectly.

**Second scenario**: One application made in **Scala** serialize the message and on the other side an other application made in **Java**,  deserialize the message and read the content.
I try to deserialize the message using the official Java module Apache Avro (https://avro.apache.org/docs/1.7.7/gettingstartedjava.html) I got an error and I am not able to deserialize the content of the message.

**Third scenario**: One application made in **Scala** serialize the message and on the other side an other application made in **Python**,  deserialize the message and read the content.
I got an error when I tried to deserialized the message (same as with Java)

**Forth scenario**: One application made in **Python** serialize the message and on the other side an other application made in **Scala**,  deserialize the message and read the content.
When I tried to deserialize the message using the avro4s module I got an error.

**Fifth scenario**: One application made in **Python** serialize the message and on the other side an other application made in **Java**,  deserialize the message and read the content.
It works perfectly!

**Sixth scenario**: One application made in **Python** serialize the message and on the other side an other application made in **JavaScript**,  deserialize the message and read the content.
It works perfectly!

I suspect there is a compatibility issue in Serializing and Deserializing messages using the avro4s and other library.

Or I'm doing something wrong ...

[...]

PR #23, opened 2016-05-17
Fixed avro binary serialization/deserialization 

annovi:
issue #21 
- removed the flag "include schema" from the AvroOutputStream because according with the Apache Avro documentation the schema is always included with the Data. So it's not need it.
[http://avro.apache.org/docs/1.8.0/gettingstartedjava.html: "Data in Avro is always stored with its corresponding schema, meaning we can always read a serialized item regardless of whether we know the schema ahead of time."
- added "AvroBinaryInputStream" and "AvroDataInputStream" that extends AvroInputStream, in order to distinguish between data file stream or binary stream
- fixed the Unit tests

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
@@ -1,24 +1,16 @@
 package com.sksamuel.avro4s
 
 import java.io.File
-import java.nio.file.Path
+import java.nio.file.{Files, Path}
 import java.io.EOFException
 
-import org.apache.avro.file.{DataFileReader, SeekableByteArrayInput, SeekableFileInput, SeekableInput}
+import org.apache.avro.file.SeekableByteArrayInput
 import org.apache.avro.generic.{GenericDatumReader, GenericRecord}
 import org.apache.avro.io.DecoderFactory
 
 import scala.util.Try
 
-trait AvroInputStream[T] {
-  def close(): Unit
-  def iterator(): Iterator[T]
-  def tryIterator(): Iterator[Try[T]]
-}
-
-class AvroBinaryInputStream[T](in: SeekableByteArrayInput)(implicit schemaFor: SchemaFor[T], fromRecord: FromRecord[T])
-  extends AvroInputStream[T] {
-
+class AvroInputStream[T](in: SeekableByteArrayInput)(implicit schemaFor: SchemaFor[T], fromRecord: FromRecord[T]) {
   val datumReader = new GenericDatumReader[GenericRecord](schemaFor())
   val binDecoder = DecoderFactory.get().binaryDecoder(in, null)
   val records = new java.util.ArrayList[GenericRecord]()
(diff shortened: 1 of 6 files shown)


sksamuel:
Thanks. I have a couple of comments.

1. Can you get all tests to pass after this PR?
2. It's allowed to not include the schema in the data. Although this is not the _normal_ case, it is supported, so I don't think we should remove the `includeSchema` flag. Maybe the flag should be renamed to `binaryMode`, or maybe people can create the different streams programatically (ie, no super trait). See https://thetechsolo.wordpress.com/2015/01/17/apache-avro-schema-less-serialization-how-to/comment-page-1/
3. Can you add a unit test for reading multiple values from a binary file, using your new Binary input stream, because I was unable to get it to work.

Thanks again.

annovi:
1) The only test that failed was the "UniverseTest.scala". Sorry,  I am going to fix it.
2) The problem with the previous solution is that when you want to serialize using **AvroBinaryOutputStream** you have to set the flag `includeSchema` to `false`. 
Let's take a look at the **AvoutOutputStream.scala** unitest:
`
   line 15:  val output = AvroOutputStream[Composer](baos, false)
`
In this example said: I want to serialize without schema (flag includeSchema = false), but then I pass the schema anyway (Composer is the schema). So I found that a bit strange.
As you suggested we can set the flag `binaryMode` to `true`, in case you want to use binary serialization. The problem with this approach (might be that I'm wrong) when you deserialize calling **AvroInputStream** we should known in advance whether to use the flag `binaryMode` to `true` or not, while most of the times you got a **serialized avro message** from an application, but you are not aware of the fact that is binary or not. If we want to deserialize a message that before was serialized as binary, we should called with a second parameter set to `true`. Anyway I open to your suggestion,  and possibile alternative solutions/implementation.
3) We do not need an additional unit test for the the version of the **AvroInputStream** because the current unit test **AvroInputStreamTest** is using feeding the class AvroInputStream with `array[Byte]` therfore if is already using the new version of the AvroBinaryInputStream that I made. The test in the  **AvroInputStreamTest** are already serializing/deserializing multiple values from a binary file.

sksamuel:
Hi,

1. Ok
2. As long as we support binary in output streams, I don't mind how its achieved. Although if we just rename the parameter to binaryMode, then it means existing code will still compile but be reversed, so perhaps that won't work. Perhaps just making both the classes directly creatable is the answer, eg
new AvroDataOutputStream(file) or whatever.

I have no problem with the fact its up to the user to know if their incoming stream has the schema embedded or not. That is something they would decide in their protocol

3. I mean a unit test that uses an input that has 2 or more case classes in a binary stream, not this:

    "read byte arrays" in {
      val data = Seq(ByteArrayTest(Array[Byte](1, 2, 3)), ByteArrayTest(Array[Byte](125, 126, 127)))
      val bytes = write(data)

      val in = AvroInputStream[ByteArrayTest](bytes)
      val result = in.iterator.toList
      result.head.bytes.toList shouldBe Array[Byte](1, 2, 3).toList
      result.last.bytes.toList shouldBe Array[Byte](125, 126, 127).toList
      in.close()
    }

Which is only testing that we can serialize byte arrays.

annovi:
- Fixed the UniverseTest.scala that was broken.
- Removed the AvroInputDataStream because is not need it.
- Restored the AvroOutputStream as it was before, with the possibility to use an additional parameter to enable/disable binary mode serialization. 

TODO: add an other Unit Test as you mention in the point 3)

annovi:
I update the pull request.

annovi:
Did you take a look at my latest PR? What do you think?

sksamuel:
I'll check it out this weekend

sksamuel:
It looks good. Only change I think is to make it so that it doesn't load all the records at the start, otherwise having an iterator is pointless and it might as well just return a Seq. I can make that change once merged in though.

annovi:
Thank you. That is great!

PR #27, opened 2016-05-23
Added Json Support for de-serialization

dlynam01:
The opposite side to the Json serialisation, allows deserialization from Json to a case class


sksamuel:
:+1:

PR #33, opened 2016-06-22
Modified ToRecord macro to cache the schema returned by schemaFor

jeroentervoorde:
Caches the schema in the ToRecord generated object. This way the same schema is always returned. This fixes a problem for us when using the confluent avro schemaregistry which does an equality check by reference. 

Besides the reduced memory allocations ofcourse.


jeroentervoorde:
In another project i am getting a compile error. Investigating first...

PR #36, opened 2016-06-28
Performance / memory usage improvements

jeroentervoorde:
This PR contains numerous improvements to avoid repeated allocations of schema's and instances of FromValue and ToValue in the macros. Furthermore is uses array based lookups instead of maps and inlines some invocations to avoid scala closure allocations.

My (basic) performance tests yields an 18X improvement in invocations ToRecord[T].apply(t) and a 2.7X improvement on FromRecord[T].apply(record)

Details:

- SchemaFor instances now store the generated schema in a local val.
- ToRecord is reuses SchemaFor instance and ToValue instances for all fields.
- ToRecord record generation loop is inlined.
- FromRecord reuses FromValue instances for all fields


sksamuel:
awesome

PR #40, opened 2016-07-09
added macro tests and data stream example

jeroentervoorde:
Added some tests which check referential equality of the generated schema's.
Added data inputstream example code to Examples test.


sksamuel:
Thanks

PR #41, opened 2016-07-18
fixed missing semicolon in enum template generator

whazenberg:
The java enum template generator wasn't including the semicolon :)


sksamuel:
:+1:

PR #42, opened 2016-07-19
handle more default values

whazenberg:
added handling of int, long, boolean, and double types


sksamuel:
This is awesome thanks

PR #44, opened 2016-07-21
default value support for array, enum, and map

whazenberg:
Added support for map and array default values

Added a test for default enum value. A default enum value is mapped to string, which is already supported.


sksamuel:
Awesome :+1:

sksamuel:
I'll release 1.5.2 now with your recent changes.

whazenberg:
Great, thanks.

PR #45, opened 2016-07-28
Fix problem with option default value

jecos:
A filed option with default value to none generate a scheam with "default : "None". Example :
`case class OptionDefaultValues(
  name: String = "sammy",
  description: Option[String] = None,
  currency: Option[String] = Some("$")
)`

produce this schema :

`{"type":"record","name":"OptionDefaultValues","namespace":"com.sksamuel.avro4s","fields":[{"name":"name","type":"string","default":"sammy"},{"name":"description","type":["null","string"],"default":"None"},{"name":"currency","type":["null","string"],"default":"Some($)"}]}`

With this fix the schema is 
`{
  "type": "record",
  "name": "OptionDefaultValues",
  "namespace": "com.sksamuel.avro4s",
  "fields": [
    { "name": "name", "type": "string", "default": "sammy" },
    {
      "name": "description",
      "type": ["null", "string"]
    },
    {
      "name": "currency",
      "type": ["null", "string"],
      "default": "$"
    }
  ]
}`

### DIFF
--- a/avro4s-core/src/test/resources/optiondefaultvalues.avsc
+++ b/avro4s-core/src/test/resources/optiondefaultvalues.avsc
@@ -1,0 +1,17 @@
+{
+  "type": "record",
+  "name": "OptionDefaultValues",
+  "namespace": "com.sksamuel.avro4s",
+  "fields": [
+    { "name": "name", "type": "string", "default": "sammy" },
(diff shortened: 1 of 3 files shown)


sksamuel:
awesome thanks

PR #46, opened 2016-08-03
Template filename bug

SohumB:
Hey! Quick little bugfix here, you were generating `domain/scala` files when using `sbt-avro4s`.


sksamuel:
awesome thanks

PR #47, opened 2016-08-04
Recursive schemas

SohumB:
Handle recursive schemas by splitting up the record's schema before fields are added, and the record's schema after fields are added, as separate values.


sksamuel:
This is brilliant

PR #48, opened 2016-08-04
targeted laziness in to/fromRecord

SohumB:
My apologies; I should've tested `toRecord` and `fromRecord` as well! 

Right now there appears to be an issue where if you're using a recursive schema, you can't rely on the `SchemaFor[T]` being implicitly built; you have to provide a `implicit val schema = SchemaFor[T]` line so that the compiler doesn't go into recursive loops. I'm tentatively okay with that; what do you think?


sksamuel:
If the only "issue" is that for recursive types, which weren't supported previously, we have to assign the schema to a variable first, then that's ok, as its still better than what we have now.

SohumB:
Makes sense to me!

PR #49, opened 2016-08-04
Coproducts!

SohumB:
Support for general union types! 

I did my best to make sure existing usecases were unimpacted; `avro4s` still uses Options and Eithers for the simpler cases, but now it can support larger unions as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,16 @@ val schema = AvroSchema[MyDecimal]
 }
 ```
 
+## Coproducts
+
+Avro supports generalised unions, eithers of more than two values. To represent these in scala, we use `shapeless.:+:`, such that `A :+: B :+: C :+: CNil` represents cases where a type is `A` OR `B` OR `C`. See shapeless' [documentation on coproducts](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#coproducts-and-discriminated-unions) for more on how to use coproducts.
+
 ## Type Mappings
+
+``` scala
+import scala.collection.{Array, List, Seq, Iterable, Set, Map, Option, Either}
+import shapeless.{:+:, CNil}
+```
 
 |Scala Type|Avro Type|
 |----------|---------|
@@ -254,14 +263,17 @@ val schema = AvroSchema[MyDecimal]
 |Java Enums|enum|
 |scala.Enumeration|enum|
 |sealed trait T|enum|
-|scala.collection.Array[T]|array|
-|scala.collection.List[T]|array|
-|scala.collection.Seq[T]|array|
-|scala.collection.Iterable[T]|array|
-|scala.collection.Set[T]|array|
-|scala.collection.Map[String, T]|map|
-|scala.collection.Option[T]|union:null,T|
-|scala.collection.Either[L, R]|union:L,R|
+|Array[T]|array|
+|List[T]|array|
+|Seq[T]|array|
+|Iterable[T]|array|
+|Set[T]|array|
+|Map[String, T]|map|
+|Option[T]|union:null,T|
+|Either[L, R]|union:L,R|
+|A :+: B :+: C :+: CNil|union:A,B,C|
+|Option[Either[L, R]]|union:null,L,R|
+|Option[A :+: B :+: C :+: CNil]|union:null,A,B,C|
 |T|record|
 
 ## Custom Type Mappings
--- a/avro4s-core/src/test/resources/optionalunion.avsc
+++ b/avro4s-core/src/test/resources/optionalunion.avsc
@@ -1,0 +1,9 @@
+{
+  "type":"record",
+  "name":"OptionalUnion",
+  "namespace":"com.sksamuel.avro4s",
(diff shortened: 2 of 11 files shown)


SohumB:
I've added a bugfix for handling of arrays inside unions. I combined it with this because the union-handling code has been touched quite a bit in this branch, but it should be cherry-pickable fairly simply into `master` as well.

SohumB:
...and similarly for maps inside unions.

SohumB:
Hey @sksamuel, I'd really appreciate it if you could say what you'd want to see from this PR to get it merged!

sksamuel:
Can you squash this so its easier to review?

So the reason I haven't merged this is because I'm a Shapeless-newb, and Shapeless is one of those very clever projects that if you don't really understand how it works, you don't really have much chance of understanding code that uses it.  I actually re-wrote the project to not use Shapeless other than `Lazy`, because it was clearer to me what was actually happening under the hood.

Can we add a bunch of comments that explain to idiots like me what the Coproducts actually do where the implicits are defined. I know Shapeless is very popular project in the pure-functional Scala community, but for people like me who aren't very advanced yet, I need to be able to still maintain the library in future :)

SohumB:
Sure, I'm happy to put in more explanation.

What exactly would you like squashed? The commits it's in currently are designed as spec/implementation pairs precisely to make them easier to review.

sksamuel:
Ok leave the commits as is then.

simonsouter:
An update to the Readme regarding the type mappings for coproducts and usage would be useful also.

SohumB:
Okay, how's that?

sksamuel:
Nice thanks. I'll merge this now.

But can you help me understand something. I see the implicits for Coproducts, and I understand about CNil being the base case, it's all very similar to HNil. But where is the magic that lets the compiler know that a type T is actually a Coproduct? I remember from using shapeless in the past that I had to have an extra step where I took a type T and used an implicit Record (or Generic?) or something to get the conversion into the HLists. What's the equivalent here for Coproduct?

SohumB:
`Coproduct` is just a sealed trait. This is the full implementation of the base types from shapeless:

```scala
sealed trait Coproduct extends Product with Serializable
sealed trait :+:[+H, +T <: Coproduct] extends Coproduct
final case class Inl[+H, +T <: Coproduct](head : H) extends :+:[H, T]
final case class Inr[+H, +T <: Coproduct](tail : T) extends :+:[H, T]
sealed trait CNil extends Coproduct {
  def impossible: Nothing
}
```

So requiring a type `A` to be `<: Coproduct` is just requiring it to be `:+:` or `CNil`; i.e., only inhabited by `Inl` and `Inr`.

sksamuel:
Yeah I know what Coproduct is an ADT of :+: or CNil, but how does my type T get converted into a Coproduct? Or does this only work if you have a Coproduct to start with ?

SohumB:
Oh! Yes, there are implicits of type `Inject[C, T]` involved. The easiest interface to them is just `Coproduct.apply`, so `Coproduct[Int :+: String :+: Boolean :+: CNil](4)`, for instance, works.

You can see some examples of that in the tests; see for instance this:

```scala
case class CoPrimitives(cp: String :+: Boolean :+: CNil)
...
type SB = String :+: Boolean :+: CNil
val data = Seq(
  CoPrimitives(Coproduct[SB]("gammy")),
  CoPrimitives(Coproduct[SB](true)),
￼  CoPrimitives(Coproduct[SB](false))
)
```

sksamuel:
Ah right ok, so the end user defines his/her type as `A :+: B :+: C`
Ok that's cool, at first I thought you were using Coproducts internally to handle sealed traits or something. I tried that and gave up.

SohumB:
Yep! The usecase here is just to be able to support larger avro unions, and coproducts are just a reasonable way to represent avro unions in Scala.

I feel like it _should_ be possible to represent sealed traits as avro unions, but that feels like it might end up exposing a lot of the issues around distinguishing between different record types that makes this code necessary:

```
// if we have a generic record, we can't use the type to work out which one it matches,
// so we have to compare field names
val typeVals: Set[String] =
tpe.members.filter(_.isTerm).map(_.asTerm).filter(_.isVal).map(_.name.decodedName.toString.trim).toSet
```

It seems to me that trying to have avro4s handle `SpecificRecord` conversions first might be better. But I'm not sure how to do that without getting into really ugly generated code with mutable vars in the case classes to implement `SpecificRecordBase`.

sksamuel:
I think with the compile time generation of the avro code, are you going to see much improvement if you switched to SpecificRecord? I find the bottleneck is always the IO.

SohumB:
I don't mean for performance; my understanding of `SpecificRecord` is that it comes with typetags so that you can distinguish a `union { RecordType, OtherRecordType }` where the two record types have the same fields.

sksamuel:
Ah right I didn't know that.

SohumB:
Though admittedly I haven't actually tested that ^_^ Just something to look into on the backburner!

PR #52, opened 2016-08-13
fix binary serialization example in README

searler:
Add corresponding tests for the examples to ensure validity


sksamuel:
Thanks1

sksamuel:
Thanks!

PR #53, opened 2016-08-13
example using custom types to map sealed trait+case object enum

searler:
Represent the following structure as Int
```
sealed trait Base
case object A extends Base
case object B extends Base
```


sksamuel:
Thanks!

Issue #55, opened 2016-08-19
BigDecimal scale and precision support


blbradley:
Hello again,

I'm looking for support for setting the scale and precision of an Avro BigDecimal. This is possible via implicits and a new type that contains the scale and precision.

I will have a contribution for it soon.

PR #56, opened 2016-08-22
set BigDecimal scale and precision for schema

blbradley:
TODOs:
* `ScaleAndPrecision` should go somewhere else. It is a big project; so, please guide me.
* Should the example print the schema of `MyDecimal`?

Ideally, it would be nice for users to have the ability to use different `ScaleAndPrecision`s within schema `BigDecimal`s. I'm not sure how to do that yet but would love to do this if it is possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,19 @@ val format = RecordFormat[Composer]
 val ennio = format.from(record)
 ```
 
+## Set a schema's decimal scale and precision
+
+Bring an implicit `ScaleAndPrecision` into scope before using `AvroSchema`.
+
+```scala
+import com.sksamuel.avro4s.ScaleAndPrecision
+
+case class MyDecimal(d: BigDecimal)
+
+implicit sp = ScaleAndPrecision(8, 20)
+val schema = AvroSchema[MyDecimal]
+```
+
 ## Type Mappings
 
 |Scala Type|Avro Type|
--- a/avro4s-macros/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
+++ b/avro4s-macros/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
@@ -27,17 +27,19 @@ trait LowPriorityToSchema {
   }
 }
 
+case class ScaleAndPrecision(scale: Int, precision: Int)
+
 object ToSchema extends LowPriorityToSchema {
 
   implicit val BooleanToSchema: ToSchema[Boolean] = new ToSchema[Boolean] {
     protected val schema = Schema.create(Schema.Type.BOOLEAN)
   }
 
-  implicit val BigDecimalToSchema: ToSchema[BigDecimal] = new ToSchema[BigDecimal] {
+  implicit def BigDecimalToSchema(implicit sp: ScaleAndPrecision = ScaleAndPrecision(2, 8)): ToSchema[BigDecimal] = new ToSchema[BigDecimal] {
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -206,6 +206,19 @@ val format = RecordFormat[Composer]
...
+## Set a schema's decimal scale and precision
+
+Bring an implicit `ScaleAndPrecision` into scope before using `AvroSchema`.
+
+```scala
+import com.sksamuel.avro4s.ScaleAndPrecision
+
+case class MyDecimal(d: BigDecimal)
+
+implicit sp = ScaleAndPrecision(8, 20)
```
sksamuel: need a `val` in there

sksamuel:
This is great thanks. I don't know how you'd have different implicits for different instances of `BigDecimal` at the same time. Might be something never really needed, can always serialise different records independently.

sksamuel:
I think that SandP can live where you put it for now. Also, it would be good for the example to show the schema as you suggest.

blbradley:
Thanks! Sorry for the typo. Closes #55.

If someone needs a more control of their `decimal` scale and precision, they can just use the Avro API directly.

SohumB:
Regarding different implicits for different BigDecimals, can you use different scopes to override implicits selectively. Something like this? (warning: untested)

```scala
case class Nested(nested: BigDecimal)
case class Data(root: BigDecimal, nested: Nested)

implicit val nestedSchema = {
  implicit val sp = ScaleAndPrecision(2,1)
  ToSchema[Nested]
}

implicit val dataSchema = {
  implicit val sp = ScaleAndPrecision(4,2)
  ToSchema[Data]
}
```

sksamuel:
I guess the issue he has that, in your example, when he makes a schema for `Data` it will include the schema for `Nested` and both will use whatever implicit is in scope.

blbradley:
@SohumB I'm afraid not at the moment. I've settled with a certain scale and precision across all my `BigDecimal`s. If you need more control, I'd use the Avro API directly.

SohumB:
I just tested it, the above code snippet works fine on current master:

```
{
  "type" : "record",
  "name" : "Data",
  "namespace" : "$iw",
  "fields" : [ {
    "name" : "root",
    "type" : {
      "type" : "bytes",
      "logicalType" : "decimal",
      "scale" : "4",
      "precision" : "2"
    }
  }, {
    "name" : "nested",
    "type" : {
      "type" : "record",
      "name" : "Nested",
      "namespace" : "$iw",
      "fields" : [ {
        "name" : "nested",
        "type" : {
          "type" : "bytes",
          "logicalType" : "decimal",
          "scale" : "2",
          "precision" : "1"
        }
      } ]
    }
  } ]
}
```

Implicits are resolved statically, and the local scopes here are distinct. So when you build `nestedSchema`, the highest priority implicit for `ScaleAndPrecision` is the `(2,1)` value, so that's what's used; and when you build `dataSchema`, the highest priority implicit _for Nested_ is `nestedSchema`, so it doesn't rebuild it. As long as `nestedSchema` is in scope when you try to build `dataSchema`, it'll work completely fine.

sksamuel:
That's great @SohumB worth adding to the readme I think.

nickpoorman:
I'm still confused on this. If I have two different decimals like the following how would you read those back in?

```
ds:org.apache.spark.sql.Dataset[org.apache.spark.sql.Row]

Field_One:decimal(19,0)
Field_Two:decimal(18,10)
```

PR #57, opened 2016-08-23
finish BigDecimal scale and precision docs

blbradley:
What you asked for. Cheers!

I used SBT's REPL. That is why the namespace is `$iw`. I don't know where that comes from.


sksamuel:
:+1:

PR #65, opened 2016-09-02
better way to discriminate between types, fixed a problem with unions...

ilya-epifanov:
better way to discriminate between types, fixed a problem with unions of types with identical field names


sksamuel:
Awesome thanks

Issue #66, opened 2016-09-12
Incompatible migration with Option


max5599:
For the case class:
`Person(name: String)`
I have the following schema:
```
{
  "type": "record",
  "name": "Person",
  "fields": [
    {
      "name": "name",
      "type": "string"
    }
  ]
}
```

I'd like to add an optional field "nickname":
`Person(name: String, nickname: Option[String] = None)`
That must result in the schema:
```
{
  "type": "record",
  "name": "Person",
  "fields": [
    {
      "name": "name",
      "type": "string"
    },
    {
      "name": "nickname",
      "type": [
        "null",
        "string"
      ],
      "default": null
    }
  ]
}
```

But unfortunately, the schema generated is missing the default to null:
```
{
  "type": "record",
  "name": "Person",
  "fields": [
    {
      "name": "name",
      "type": "string"
    },
    {
      "name": "nickname",
      "type": [
        "null",
        "string"
      ]
    }
  ]
}
```
And so it's not compatible.

PR #67, opened 2016-09-12
Set default to null for Option with default None

max5599:
Related to Incompatible migration with Option #66


sksamuel:
Awesome thanks.

Issue #76, opened 2016-11-07
Getting warning of local val never used, using AvroInputStream.data


RayRoestenburg:
When using `AvroInputStream.data[SomeType](bytes)` I'm getting a warning, 

```local val in value x$1 is never used``` 

I use `-Xfatal-warnings` as scalac option, which means this turns into an error. 
Am I doing something wrong here, or do you ignore warnings?

The below code triggers the warning:

```
package my.pckg

import com.sksamuel.avro4s._
case class Bla(name: String)
object Avro {
  def fromAvro(bytes: Array[Byte]) = {
    AvroInputStream.data[Bla](bytes)
  }
}
```

markschaake:
I'm also seeing this error, and also have `-Xfatal-warnings` as a scalac option.

PR #77, opened 2016-11-08
Remove all compiler warnings

markschaake:
This PR removes all compiler warnings and requires going forward that no more warnings are allowed by adding `-Xfatal-warnings` to the build's `scalacOptions`.

This PR was motivated by #76, though it does not fix the bug. I am trying to reproduce #76 via a unit test but so far have not been able to. However, please consider merging these small changes to prevent compiler warnings being introduced in the future.

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/AvroInputStreamTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/AvroInputStreamTest.scala
@@ -2,13 +2,12 @@ package com.sksamuel.avro4s
 
 import java.io.ByteArrayOutputStream
 import java.util.UUID
-
-import org.scalatest.concurrent.Timeouts
+import org.scalatest.concurrent.TimeLimits
 import org.scalatest.{Matchers, WordSpec}
 
 import shapeless.{:+:, Coproduct, CNil}
 
-class AvroInputStreamTest extends WordSpec with Matchers with Timeouts {
+class AvroInputStreamTest extends WordSpec with Matchers with TimeLimits {
 
   case class Booleans(bool: Boolean)
   case class BigDecimalTest(decimal: BigDecimal)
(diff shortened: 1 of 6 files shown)


avro4s-core/src/test/scala/com/sksamuel/avro4s/AvroInputStreamTest.scala
```diff
@@ -2,13 +2,12 @@ package com.sksamuel.avro4s
...
 import java.util.UUID
-
-import org.scalatest.concurrent.Timeouts
+import org.scalatest.concurrent.TimeLimits
 import org.scalatest.{Matchers, WordSpec}
 
 import shapeless.{:+:, Coproduct, CNil}
 
-class AvroInputStreamTest extends WordSpec with Matchers with Timeouts {
+class AvroInputStreamTest extends WordSpec with Matchers with TimeLimits {
```
markschaake: `Timeoutes` was deprectated in favor of `TimeLimits`

avro4s-json/src/main/scala/com/sksamuel/avro4s/json/JsonToAvroConverter.scala
```diff
@@ -38,7 +38,9 @@ class JsonToAvroConverter(namespace: String, avroStringTypeIsString: Boolean = f
     case JString(_) => createStringSchema
     case JObject(values) =>
       val record = Schema.createRecord(name, null, namespace, false)
-      val fields = values.map { case (name, value) => new Schema.Field(name, convert(name, value), null, null) }
+      val doc: String = null
+      val default: AnyRef = null
+      val fields = values.map { case (name, value) => new Schema.Field(name, convert(name, value), doc, default) }
```
markschaake: Some `Schema.Field` constructors were deprecated. Providing explicit types for the arguments ensures the correct non-deprecated constructor is called.

avro4s-macros/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
```diff
@@ -370,28 +367,21 @@ object SchemaFor {
...
+    def toDefaultValue(value: Any): Any = value match {
+      case x: Int => x
+      case x: Long => x
+      case x: Boolean => x
+      case x: Double => x
+      case x: Seq[_] => x.asJava
+      case x: Map[_, _] => x.asJava
+      case Some(x) => x
+      case None => JsonProperties.NULL_VALUE
+      case _ => value.toString
```
markschaake: This is the riskiest change in the PR. The `Schema.Field` constructors that took a `JsonNode` argument for the default value have all been deprecated. There was a unit test in place that covered the expected behavior here, and it still passes with this change.

sksamuel:
:+1:

PR #87, opened 2016-12-08
proper fixed type support

ilya-epifanov:
Associated PR: https://github.com/sksamuel/sbt-avro4s/pull/9

Added proper support for `fixed` types, currently represented as:
`@AvroFixed(32) case class SHA256(bytes: WrappedArray.ofByte) extends AnyVal`


sksamuel:
:+1:

PR #88, opened 2016-12-10
Add support for primitives

mtranter:
Hi!

Using the current implementation the following is not possible,

  val boas = new ByteArrayOutputStream()
  val output = AvroOutputStream.binary[String](boas)
  output.write("hello avro")

Because internally, a GenericDatumReader/Write of type [GenericRecord] is being used and primitives arent avro record types.
If we switch this out for a GenericDatumReader/Writer of type [Any], and switch from using implicit From/ToRecord[T] to using From/ToValue[T] in the public API, we get the ability to work with primitives. We also dont loose any functionality around GenericRecord/complex types (due to the implicit conversions between the To/FromValue type classes to the To/FromRecord type classes).

This is a risky change, as it changes the public API, however due to the implict conversions mentioned above I can't imagine there being any breaking changes. (It was also a trivially easy change thanks to the awesome code!)

I've submitted this more for discussion than actually expecting it to be merged as is. I think the ability to use primitives directly (rather than wrapping them in case classes) is a key use case.

Any thoughts??

Thanks

Mark.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
@@ -16,7 +16,7 @@ trait AvroInputStream[T] {
 }
 
 class AvroBinaryInputStream[T](in: InputStream)
-                              (implicit schemaFor: SchemaFor[T], fromRecord: FromRecord[T])
+                              (implicit schemaFor: SchemaFor[T], fromValue: FromValue[T])
   extends AvroInputStream[T] {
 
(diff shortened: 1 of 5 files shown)


sksamuel:
Seems a reasonable change. I'll take a look over the code when I get chance.

sksamuel:
I think this looks good as it is.

mtranter:
Great. Any chance it can be merged in?

LukeFieldsend:
I need this too, would be great to push it through

moritzhaller:
Any updates on this? Would need this too!

sksamuel:
This has been added manually, and the PR should be closed. Thank you for your PR, and sorry it's taken so long to acknowledge it.

PR #94, opened 2017-01-12
Removed support for scala.Enumeration

YuvalItzchakov:
scala.Enumeration is not supported by avro4s, this is an update to let all users know.


sksamuel:
Thanks

Issue #100, opened 2017-02-26
Can not enclose case classes inside lowercase only named object


LGLO:
Package name heuristic fails when schema case classes are placed in lowercase named object.
This is minimal example I can provide.
```scala
object lowercase extends App {

  import com.sksamuel.avro4s.{FromRecord, ToRecord}
  import shapeless.{:+:, CNil, Coproduct}

  case class A(a: Int)
  case class B(b: String)
  type AOrB = A :+: B :+: CNil
  case class AB(ab: AOrB)

  val from = FromRecord[AB]
  val to = ToRecord[AB]

  from(to(AB(Coproduct[AOrB](A(0)))))
}
```

Changing package name to `Uppercase` would fix application.

LGLO:
Thank you!

PR #101, opened 2017-02-26
Real package path in generated schema

krzemin:
Real package path in generated schema retrieved in generated schema instead of lowercase heuristics. It addresses #100 by unifying method of determining package name of a type both in `FromValue` typeclass derivation for coproducts and generation of schema in `SchemaFor` macro.


sksamuel:
awesome

PR #102, opened 2017-02-26
Introduced consistent end-to-end, union-based encoding of sealed traits

krzemin:
Instead of custom encoding for sealed traits, introduced encoding as unions using shapeless.Generic. Kept optimal, string-based encoding in case when all sealed hierarchy descendants are singleton objects.

@sksamuel I know this is breaking change, but brings nice support for end-to-end sealed hierarchies. Please tell what do you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,13 @@ val schema = AvroSchema[MyDecimal]
 
 Avro supports generalised unions, eithers of more than two values. To represent these in scala, we use `shapeless.:+:`, such that `A :+: B :+: C :+: CNil` represents cases where a type is `A` OR `B` OR `C`. See shapeless' [documentation on coproducts](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#coproducts-and-discriminated-unions) for more on how to use coproducts.
 
+## Sealed hierarchies
+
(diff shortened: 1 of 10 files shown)


sksamuel:
I don't think we should force people to use shapeless unless they want to.

krzemin:
Are you suggesting just to rewrite the same with macro?

sksamuel:
No sorry. I thought the user would have to use coproducts in their domain code. I'm happy for shapeless to be used internally.

krzemin:
Well, actually that change aims to free user from using coproducts in domain code :)

sksamuel:
I'll merge it later as I want to look it over on my computer first to make
sure I understand it. I'm just on a phone atm.

krzemin:
I updated the docs.

sksamuel:
I wonder if we should change sealed traits + case objects to be an avro enum ?

krzemin:
Thanks for merging.

Semantic meaning of such a sealed trait + case objects is mostly for having more type safe enumerations. I think it may be possible to encode them as avro enums, with little effort. Will make another PR soon.

sksamuel:
Should be, just a matter of changing the schema type when its using the genobject method.

PR #103, opened 2017-02-27
Sealed trait + case object encoded as enum

krzemin:
Following discussion in #102, replaced string-based encoding for sealed trait + only case object descendants with enum-based one.


krzemin:
Thanks. Any chance for release soon?

PR #109, opened 2017-03-13
Handle generics when type is fully known

nightscape:
This is a first shot at handling generics.
I have only added a simple testcase, but it might well work for more complicated ones.
It would be great to get generic handling in some form into the next release! :smiley:

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/RecordFormatTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/RecordFormatTest.scala
@@ -5,6 +5,11 @@ import shapeless.{Inl, Inr}
 
 class RecordFormatTest extends WordSpec with Matchers {
   case class Composer(name: String, birthplace: String, compositions: Seq[String])
+
+  case class Book(title: String)
+  case class Song(title: String, lyrics: String)
+  case class Author[T](name: String, birthplace: String, work: Seq[T])
+
   "RecordFormat" should {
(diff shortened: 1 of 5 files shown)


nightscape:
For people interested in experimenting with this, you can try the Jitpack build of our fork.
Just add the following to your `build.sbt`
```scala
resolvers += "jitpack" at "https://jitpack.io"
libraryDependencies += "com.github.crealytics" % "avro4s" % "handle_generics-SNAPSHOT"

```

nightscape:
Ok, I discovered a first caveat: The following will currently not work because of a name clash of the differently typed `Generic`s:
```scala
case class Generic[T](t: T)
case class SameGenericWithDifferentTypeArgs(gi: Generic[Int], gs: Generic[String])
val schema = SchemaFor[SameGenericWithDifferentTypeArgs]()
```
Does anyone know, would it be legal to name the types `Generic[Int]` and `Generic[String]` in the schema?

nightscape:
Ok, according to the official Avro specs, only letters, numbers and underscores are allowed:
https://avro.apache.org/docs/1.8.1/spec.html#names
@sksamuel Do you know if the name generation has to be invertible? I.e. is the name parsed somewhere in order to get the class name?

sksamuel:
Not that I can remember. I think you're OK to do something like
List__String__ to encode it.

sksamuel:
Your work in this PR has been manually added to the 2.0.0 branch, so I will close this PR now. Thank you.

Issue #113, opened 2017-03-23
Binary reader API does't support schema evolution.


NetanelRabinowitz:
Avro has a GenericData.createDatumReader(Schema writer, Schema reader) API
Avro4s abstract over this API but if we want to read a record via the binary API `AvroInputStream.binary[newClass](in)`
we lost the ability of schema evolution because currently there is no way to pass the writer schema.
I think I't would be super beneficial to add support for this kind of API and enable schema evolution.

OmriManor:
+1
A very valid use case, for example, when having multiple binary records in kafka, one should be able to deserialize the old/new records with a readers and writers schema.

I considered using avro4s and this was the feature that was missing.

Are there future plans on adding this support?

PR #114, opened 2017-03-23
support binary schema evolution

NetanelRabinowitz:
This is a fix for issue #113

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
@@ -16,7 +16,7 @@ trait AvroInputStream[T] {
   def tryIterator(): Iterator[Try[T]]
 }
 
-class AvroBinaryInputStream[T](in: InputStream, writerSchema: Option[Schema])
+class AvroBinaryInputStream[T](in: InputStream, writerSchema: Option[Schema] = None)
                               (implicit schemaFor: SchemaFor[T], fromRecord: FromRecord[T])
   extends AvroInputStream[T] {
   val wSchema = writerSchema.getOrElse(schemaFor())
@@ -103,7 +103,7 @@ object AvroInputStream {
   def binary[T: SchemaFor : FromRecord](path: Path, writerSchema: Schema): AvroBinaryInputStream[T] = binary(path.toFile, writerSchema)
 
   // convenience api for cases where the writer schema should be the same as the reader.
-  def binary[T: SchemaFor : FromRecord](in: InputStream): AvroBinaryInputStream[T] = new AvroBinaryInputStream[T](in, None)
+  def binary[T: SchemaFor : FromRecord](in: InputStream): AvroBinaryInputStream[T] = new AvroBinaryInputStream[T](in)
   def binary[T: SchemaFor : FromRecord](bytes: Array[Byte]): AvroBinaryInputStream[T] = binary(new SeekableByteArrayInput(bytes))
   def binary[T: SchemaFor : FromRecord](file: File): AvroBinaryInputStream[T] = binary(new SeekableFileInput(file))
   def binary[T: SchemaFor : FromRecord](path: String): AvroBinaryInputStream[T] = binary(Paths.get(path))
@@ -115,7 +115,7 @@ object AvroInputStream {
   def data[T: SchemaFor : FromRecord](path: Path): AvroDataInputStream[T] = data(path.toFile)
 
   @deprecated("Use .json .data or .binary to make it explicit which type of output you want", "1.5.0")
-  def apply[T: SchemaFor : FromRecord](bytes: Array[Byte]): AvroInputStream[T] = new AvroBinaryInputStream[T](new SeekableByteArrayInput(bytes), None)
+  def apply[T: SchemaFor : FromRecord](bytes: Array[Byte]): AvroInputStream[T] = new AvroBinaryInputStream[T](new SeekableByteArrayInput(bytes))
   @deprecated("Use .json .data or .binary to make it explicit which type of output you want", "1.5.0")
   def apply[T: SchemaFor : FromRecord](path: String): AvroInputStream[T] = apply(new File(path))
   @deprecated("Use .json .data or .binary to make it explicit which type of output you want", "1.5.0")
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
@@ -3,6 +3,7 @@ package com.sksamuel.avro4s
 import java.io.{EOFException, File, InputStream}
 import java.nio.file.{Path, Paths}
 
+import org.apache.avro.Schema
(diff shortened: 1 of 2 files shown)


avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
```diff
@@ -15,11 +15,11 @@ trait AvroInputStream[T] {
   def tryIterator(): Iterator[Try[T]]
 }
 
-class AvroBinaryInputStream[T](in: InputStream)
+class AvroBinaryInputStream[T](in: InputStream, writerSchema: org.apache.avro.Schema = null)
                               (implicit schemaFor: SchemaFor[T], fromRecord: FromRecord[T])
   extends AvroInputStream[T] {
-
-  val datumReader = new GenericDatumReader[GenericRecord](schemaFor())
+  val wSchema = if (writerSchema == null) schemaFor() else writerSchema
```
NetanelRabinowitz: I don't like this null check but didn't want to break the API.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
```diff
@@ -101,6 +101,12 @@ object AvroInputStream {
   def binary[T: SchemaFor : FromRecord](path: String): AvroBinaryInputStream[T] = binary(Paths.get(path))
   def binary[T: SchemaFor : FromRecord](path: Path): AvroBinaryInputStream[T] = binary(path.toFile)
 
+  def binaryEvolve[T: SchemaFor : FromRecord](in: InputStream, writerSchema: org.apache.avro.Schema): AvroBinaryInputStream[T] = new AvroBinaryInputStream[T](in, writerSchema)
```
NetanelRabinowitz: Ideally the writerSchema has the same schema as `SchemaFor[T].apply()` as default value on `binary`  but we can't overload with default values :/

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
```diff
@@ -95,7 +96,14 @@ object AvroInputStream {
   def json[T: SchemaFor : FromRecord](path: String): AvroJsonInputStream[T] = json(Paths.get(path))
   def json[T: SchemaFor : FromRecord](path: Path): AvroJsonInputStream[T] = json(path.toFile)
 
-  def binary[T: SchemaFor : FromRecord](in: InputStream): AvroBinaryInputStream[T] = new AvroBinaryInputStream[T](in)
+  def binary[T: SchemaFor : FromRecord](in: InputStream, writerSchema: Schema): AvroBinaryInputStream[T] = new AvroBinaryInputStream[T](in, Option(writerSchema))
```
NetanelRabinowitz: Ideally the writerSchema has the same schema as AvroSchema[T] as a default value on binary but we can't overload with default values :/

NetanelRabinowitz:
@sksamuel any thoughts?

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
```diff
@@ -15,11 +15,11 @@ trait AvroInputStream[T] {
   def tryIterator(): Iterator[Try[T]]
 }
 
-class AvroBinaryInputStream[T](in: InputStream)
+class AvroBinaryInputStream[T](in: InputStream, writerSchema: org.apache.avro.Schema = null)
```
sksamuel: Maybe we're better off breaking the API and having the input stream always need a writer, of which it can be provided via an implicit on an apply method?

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
```diff
@@ -101,6 +101,12 @@ object AvroInputStream {
   def binary[T: SchemaFor : FromRecord](path: String): AvroBinaryInputStream[T] = binary(Paths.get(path))
   def binary[T: SchemaFor : FromRecord](path: Path): AvroBinaryInputStream[T] = binary(path.toFile)
 
+  def binaryEvolve[T: SchemaFor : FromRecord](in: InputStream, writerSchema: org.apache.avro.Schema): AvroBinaryInputStream[T] = new AvroBinaryInputStream[T](in, writerSchema)
+  def binaryEvolve[T: SchemaFor : FromRecord](bytes: Array[Byte], writerSchema: org.apache.avro.Schema): AvroBinaryInputStream[T] = binaryEvolve(new SeekableByteArrayInput(bytes), writerSchema)
+  def binaryEvolve[T: SchemaFor : FromRecord](file: File, writerSchema: org.apache.avro.Schema): AvroBinaryInputStream[T] = binaryEvolve(new SeekableFileInput(file), writerSchema)
+  def binaryEvolve[T: SchemaFor : FromRecord](path: String, writerSchema: org.apache.avro.Schema): AvroBinaryInputStream[T] = binaryEvolve(Paths.get(path), writerSchema)
+  def binaryEvolve[T: SchemaFor : FromRecord](path: Path, writerSchema: org.apache.avro.Schema): AvroBinaryInputStream[T] = binaryEvolve(path.toFile, writerSchema)
```
sksamuel: Does it make sense to add these for the data input stream too?

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
NetanelRabinowitz: There is a [later commit](https://github.com/sksamuel/avro4s/pull/114/commits/960b5b3d398126e96bc923261329951c338fc227) where I update the API, 
I't let you specify the writer schema on the binary method, (not binaryEvolve).
it keeps the API nice and clean on the user site.
implicit here for the writer schema is a bit problematic because the common use case is when the same class was evolve, it now different from the time of writing.
so we can do our best by letting the user pass the actual avro schema he pulled from a schema registry or something.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
NetanelRabinowitz: I don't think so because when reading from a file you have the writer schema encodes within the file itself so on the call site the user doesn't need to specify it..

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
NetanelRabinowitz: I actually start by writing this version of the api that provide the schema via implicit. but than I realize, that if the class change over time I just lost the writer schema :)

NetanelRabinowitz:
@sksamuel I answered on the outdated commit, don't know if you saw it.. anyway can you take a look at the updated PR? Tell me what you think..

sksamuel:
I think its good. Can we make the constructor use a default for the schema, so anyone who is doing `new Stream` won't be affected.

NetanelRabinowitz:
@sksamuel yea good idea! I've just updated the PR.

sksamuel:
:+1:

NetanelRabinowitz:
Thanks !!

PR #122, opened 2017-05-10
Add support for modifying namespace in record types

YuvalItzchakov:
For record types, it is essential to be able to modify namespaces of records, especially if you're avro is meant to be read by multiple languages on different platforms.

This pull request adds the ability to annotate a case class with @AvroNamespace to modify the generated namespace name in the schema.

For example:

```scala
package com.github.yuval
@AvroNamespace("com.yuval") case class Foo(s: String)
```

Will generate:

```json
{
    "type": "record",
    "name": "Foo",
    "namespace": "com.yuval",
    "fields": [
        {
            "name": "s",
            "type": "string"
        }
    ]
}
```

### DIFF
--- a/avro4s-core/src/test/resources/namespace.avsc
+++ b/avro4s-core/src/test/resources/namespace.avsc
@@ -1,0 +1,20 @@
+{
+  "type" : "record",
+  "name" : "AnnotatedNamespace",
+  "namespace" : "com.yuval",
+  "fields" : [ {
+    "name" : "s",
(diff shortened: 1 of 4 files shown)


sksamuel:
:+1:

MLnick:
Looks like this would be necessary for me too. When is the next release likely?

sksamuel:
I will release 1.7 within the next week.

MLnick:
Thanks!

PR #123, opened 2017-05-17
Avoid unnecessary string allocations in `FromValue` implicits

YuvalItzchakov:
When one of the `FromValue` implicits for primitive types parses the actual type of the value, it first allocates a string and then converts to the primitive type:

For example:

```scala
implicit object IntFromValue extends FromValue[Int] {
  override def apply(value: Any, field: Field): Int = value.toString.toInt
}
```

Under heavy load, this generates a decent amount of garbage strings. As an example, here is a sample from a profile I ran on code which consumes ~ 500,000 avro records:

Sample profile on Avro heavy app

We generate nearly 1GB of strings for parsing out `Int`s.

Since [`safeFrom`](https://github.com/sksamuel/avro4s/blob/master/avro4s-macros/src/main/scala/com/sksamuel/avro4s/FromRecord.scala#L143) matches type parameters via `WeakTypeTag`, we can safely cast to the underlying primitive type without needs to allocate an additional string.

This pull request uses `asInstanceOf` for all primitive types instead of `toString`.

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/RecordFormatTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/RecordFormatTest.scala
@@ -5,6 +5,11 @@ import shapeless.{Inl, Inr}
 
 class RecordFormatTest extends WordSpec with Matchers {
   case class Composer(name: String, birthplace: String, compositions: Seq[String])
+
+  case class Book(title: String)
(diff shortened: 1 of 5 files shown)


YuvalItzchakov:
Samuel?

sksamuel:
Looks very good. Sorry for the late response.

Issue #130, opened 2017-06-20
Using different schema for reading and writing


piotr-kalanski:
After running below code I have exception:

```scala
// Sample model:
  case class PersonV2(name: String, age: Int, title: Option[String])
  case class PersonV3(name: String, age: Int, title: Option[String], salary: Option[Long])

  val peopleV2 = Seq(
    PersonV2("p1", 10, Some("Mr")),
    PersonV2("p2", 20, Some("Ms")),
    PersonV2("p3", 30, None),
    PersonV2("p4", 40, Some("Mr"))
  )

// Saving version 2:
  val file = new File("target/peoplev2.avro")
  val os = AvroOutputStream.data[PersonV2](file)
  os.write(peopleV2)
  os.close()

// Saving version 3 which is compatible with version 2:
  val peopleV3is = AvroInputStream.data[PersonV3](file)
  val peopleV3 = Seq(
    PersonV3("p1", 10, Some("Mr"), None),
    PersonV3("p2", 20, Some("Ms"), None),
    PersonV3("p3", 30, None, None),
    PersonV3("p4", 40, Some("Mr"), None)
  )
```

Exception:
```
Exception in thread "main" org.apache.avro.AvroTypeException: Found com.sksamuel.avro4s.PersonV2, expecting com.sksamuel.avro4s.PersonV3, missing required field salary
	at org.apache.avro.io.ResolvingDecoder.doAction(ResolvingDecoder.java:292)
	at org.apache.avro.io.parsing.Parser.advance(Parser.java:88)
	at org.apache.avro.io.ResolvingDecoder.readFieldOrder(ResolvingDecoder.java:130)
	at org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:223)
	at org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:174)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:152)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:144)
	at org.apache.avro.file.DataFileStream.next(DataFileStream.java:233)
	at org.apache.avro.file.DataFileStream.next(DataFileStream.java:220)
	at com.sksamuel.avro4s.AvroDataInputStream$$anon$3.next(AvroInputStream.scala:57)
	at scala.collection.Iterator.foreach(Iterator.scala:929)
	at scala.collection.Iterator.foreach$(Iterator.scala:929)
[...]

PR #131, opened 2017-06-20
Using different schema for reading and writing

piotr-kalanski:
My proposition for solving issue described in ticket: https://github.com/sksamuel/avro4s/issues/130

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
@@ -49,7 +49,7 @@ class AvroDataInputStream[T](in: SeekableInput)
                             (implicit schemaFor: SchemaFor[T], fromRecord: FromRecord[T])
   extends AvroInputStream[T] {
 
-  val datumReader = new GenericDatumReader[GenericRecord](schemaFor())
+  val datumReader = new GenericDatumReader[GenericRecord]()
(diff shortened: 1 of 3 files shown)


sksamuel:
I like the change, but if we remove the schemaFor implicit, then users cannot provide their own schema anymore, but have to use a case class to model the required schema. Which is nice, that it's strongly typed, but I think a previous request was to allow users to specify the schema for reading manually. So perhaps we need to support both use cases?

`AvroDataInputStream.forReadSchema[TypeOfSchema]` or something ?

piotr-kalanski:
Hi,

I've added new API for constructing `AvroInputSchema` using Builder pattern, where you can specify format, source, reader schema and writer schema. I prefer such API, because currently in `AvroInputStream` companion object there are already many methods and adding methods for all possible combinations is not convenient :) I've also a little changed Json and Binary input streams to allow providing reader and writer schema. There are also many new unit tests to test new API. Please let me know what do you think about it. If you like it we can also rebuild in similar way `AvroOutputStream` API.

I have only problem with adding support for reading with new version of reader for Binary and Json formats, because of method `GenericDatumReader.read` which requires not null reader schema and after providing schema for new version there is exception related to missing data for new field even if it is optional... Probably to add support for this it will require using low level `Avro` API based on class: `org.apache.avro.file.DataFileReader` which is used in `AvroDataInputStream`.

Regards,
Piotr

sksamuel:
Looks great!

PR #132, opened 2017-06-26
Inject schema in derivation

fydio:
When a custom `SchemaFor` is used it might diverge from the ones derived in `ToRecord` and `ToValue`.

With this PR it is possible to supply the manually created ones so they are used in the `ToRecord` and `ToValue` derivation.

### DIFF
--- a/avro4s-macros/src/main/scala/com/sksamuel/avro4s/ToRecord.scala
+++ b/avro4s-macros/src/main/scala/com/sksamuel/avro4s/ToRecord.scala
@@ -163,6 +163,7 @@ trait ToRecord[T] extends Serializable {
 object ToRecord {
 
   implicit def apply[T]: ToRecord[T] = macro applyImpl[T]
+  def withSchemaFor[T](schemaFor: SchemaFor[T]): ToRecord[T] = macro withSchemaForImpl[T]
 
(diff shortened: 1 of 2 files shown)


avro4s-macros/src/main/scala/com/sksamuel/avro4s/ToRecord.scala
```diff
@@ -27,13 +27,23 @@ trait LowPriorityToValue {
   }
 
   def fixed[T]: ToValue[T] = macro LowPriorityToValue.fixedImpl[T]
```
fydio: Not sure if this is still in use.

avro4s-macros/src/main/scala/com/sksamuel/avro4s/ToRecord.scala
```diff
@@ -27,13 +27,23 @@ trait LowPriorityToValue {
...
+
+  def fixedWithSchemaFor[T](schemaFor: SchemaFor[T]): ToValue[T] = macro LowPriorityToValue.withSchemaForImpl[T]
 }
 
 object LowPriorityToValue {
   def fixedImpl[T: c.WeakTypeTag](c: scala.reflect.macros.whitebox.Context): c.Expr[ToValue[T]] = {
```
fydio: Not sure if this is still used either.

fydio:
The need for this came up due to incorrect usage of the library. Closing this.

PR #143, opened 2017-09-05
Add support for java LocalDate type

NicolaeNMV:
Hello there,

I thought it may be useful to have support for the java LocalDate inside avro4s.

It may be a good idea to start grouping similar types in a coherent modules `TimeInstances` ? I was looking how circe handled that: 
https://github.com/circe/circe/blob/master/modules/java8/src/main/scala/io/circe/java8/time/TimeInstances.scala
Also another idea to consider would be to have the definition for all the implicits SchemaFor/FromValue/ToValue for a given type in one place rather than having them split in different files by implicit type as it's done now.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,7 @@ import shapeless.{:+:, CNil}
 |Double|double|
 |Float|float|
 |java.util.UUID|string|
+|java.time.LocalDate|string|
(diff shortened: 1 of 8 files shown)


sksamuel:
👍

PR #147, opened 2017-09-21
BigDecimalToValue should use a RoundingMode when scaling a BigDecimal…

xelnaga:
Applying a scale to BigDecimal values will throw an exception if the number needs to be rounded unless a rounding mode is also specified.


jeantil:
Shouldn't you use `BigDecimal.RoundingMode.UNNECESSARY` by default instead of `BigDecimal.RoundingMode.HALF_EVEN`

Your current fix will lead to silent dataloss which can be fairly hard to diagnose. Yes an exception is bad but at least you know something is wrong and can decide on a case by case whether to extend the schema precision or to round the value...

xelnaga:
Seems reasonable.  I've updated the PR.

sksamuel:
:+1:

Issue #135, opened 2017-07-06
NPE when writing Option of sealed trait hierarchy of case objects


jangoolie:
This test demonstrates the issue:
```scala

sealed trait EnumThing
case object Thing extends EnumThing
case object Thang extends EnumThing
case class MaybeEnumThing(eo: Option[EnumThing])

...

"be able to write a non-null union of an enum and null" in {
  val fmt = RecordFormat[MaybeEnumThing]

  val met0 = MaybeEnumThing(Some(Thing))
  val baos = new ByteArrayOutputStream()
  val output = AvroOutputStream.binary[MaybeEnumThing](baos)
  output.write(met0) //<------ Throws a NullPointerException
  output.close()

  val is = AvroInputStream.binary[MaybeEnumThing](baos.toByteArray)
  val met1 = is.iterator.toList.head
  is.close()

  met0 shouldEqual met1
}   
```
Causes a NPE to be thrown (stack trace below).

Note that 
```scala
val met0 = MaybeEnumThing(None)
``` 
Works fine. So this bug only manifests when writing a non-null value to a union of null and Enum.

[info] - should be able to write a non-null union of an enum and null *** FAILED ***
[info]   java.lang.NullPointerException: null of union in field eo of com.sksamuel.avro4s.MaybeEnumThing
[info]   at org.apache.avro.generic.GenericDatumWriter.npe(GenericDatumWriter.java:132)
[info]   at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:126)
[info]   at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:73)
[info]   at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:60)
[info]   at com.sksamuel.avro4s.AvroBinaryOutputStream.write(AvroOutputStream.scala:32)
[info]   at com.sksamuel.avro4s.AvroBinaryTest.$anonfun$new$5(AvroBinaryTest.scala:72)
[info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   ...
[...]

PR #158, opened 2017-11-02
Fix for issue #135: NPE on Option[Enumeration]

svroonland:
Fixes issue #135 

My knowledge of the reflection api is limited, so the solution may not be optimal.

Also note there may be similar cases, where some types `T` are not supported within an Option[T].


sksamuel:
:+1:

Issue #110, opened 2017-03-14
RecordFormat[T].from(...) not picking up default values of schema


pjan:
How to reproduce:

```
case class P1(name: String, age: Int = 18)
case class P2(name: String)

val f1 = RecordFormat[P1]
val f2 = RecordFormat[P2]

f1.from(f2.to(P2("test")))
```

This blows up in a nullpointer exception during runtime. Expected behaviour would be that the default value of the schema gets picked up, and the code doesn't blow up. Or am I missing something?

sksamuel:
Try making the formats implicit.

pjan:
Not sure how this would solve anything, since nothing in the code example requires implicit `formats`.

Anyway, tried making them implicit, and it doesn't change anything, unfortunately...

sksamuel:
Never mind, misunderstood.

pjan:
It does work when I use output & inputstreams, so I am probably mis-using the api in the example above. Can close if that would be the case/suspicion.

Bhashit:
Not to push or anything, but is anyone working on this? If not, would you mind if I took this up? I am getting the same issue.

sksamuel:
@Bhashit please feel free to tackle it

sksamuel:
This is fixed in 2.0.0-M3 onwards.

See test using your code added here:
https://github.com/sksamuel/avro4s/blob/master/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue110.scala

PR #167, opened 2018-01-21
WIP: Allow optionally specifying a reader-schema when creating FromRecords

Bhashit:
Issue #110

Allow optionally specifying a reader-schema when creating FromRecords. Also, add `null` handling for `FromValue` instances (WIP).

@sksamuel I have added some initial changes. The macro related changes are working as expected. I ran some tests. If you don't mind, can you please take a look at how I have handled `null` values for some of the `FromValue` instances here. If you think I'm going in the right direction, I can make similar changes elsewhere.

### DIFF
--- a/avro4s-macros/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
+++ b/avro4s-macros/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
@@ -8,7 +8,7 @@ import com.sksamuel.avro4s.ToSchema.defaultScaleAndPrecision
 import org.apache.avro.Schema.Field
 import org.apache.avro.generic.{GenericData, GenericRecord}
 import org.apache.avro.util.Utf8
-import org.apache.avro.{Conversions, LogicalTypes}
+import org.apache.avro.{Conversions, LogicalTypes, Schema}
 import shapeless.ops.coproduct.Reify
 import shapeless.ops.hlist.ToList
 import shapeless.{:+:, CNil, Coproduct, Generic, HList, Inr, Lazy}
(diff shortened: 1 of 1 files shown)


Bhashit:
@sksamuel just pinging here to find out if this can be still be considered for getting in. Or, if I need to make some other changes.

sksamuel:
Sorry I wasn't looking at it until the WIP: was removed. I'll examine it later and feed back.

Bhashit:
Oh, great. Thank you.

sksamuel:
I don't really understand what you've done here. There seems to be use of default values. Can you explain what this has to do with reader schema ?

avro4s-macros/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
```diff
@@ -45,7 +45,11 @@ object FromValue extends LowPriorityFromValue {
       val decimalConversion = new Conversions.DecimalConversion
       val decimalType = LogicalTypes.decimal(sp.precision, sp.scale)
       override def apply(value: Any, field: Field): BigDecimal = {
-        decimalConversion.fromBytes(value.asInstanceOf[ByteBuffer], null, decimalType)
+        if (value == null && field.defaultVal() != null) {
+          decimalConversion.fromBytes(ByteBuffer.wrap(field.defaultVal.asInstanceOf[Array[Byte]]), null, decimalType)
+        } else {
+          decimalConversion.fromBytes(value.asInstanceOf[ByteBuffer], null, decimalType)
+        }
       }
```
sksamuel: Do you think the default value should be used on a read, or just a write?

avro4s-macros/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
```diff
@@ -45,7 +45,11 @@ object FromValue extends LowPriorityFromValue {
       val decimalConversion = new Conversions.DecimalConversion
       val decimalType = LogicalTypes.decimal(sp.precision, sp.scale)
       override def apply(value: Any, field: Field): BigDecimal = {
-        decimalConversion.fromBytes(value.asInstanceOf[ByteBuffer], null, decimalType)
+        if (value == null && field.defaultVal() != null) {
+          decimalConversion.fromBytes(ByteBuffer.wrap(field.defaultVal.asInstanceOf[Array[Byte]]), null, decimalType)
+        } else {
+          decimalConversion.fromBytes(value.asInstanceOf[ByteBuffer], null, decimalType)
+        }
```
sksamuel: Do you think the default value should be used on a read, or just a write?

Bhashit: I think it should be for reading as well. I mean, if the value doesn't exist in the record, and there's a default in the schema, it may make sense to get the default.

Bhashit:
> I don't really understand what you've done here. There seems to be use of default values. Can you explain what this has to do with reader schema ?

I'll try to put together an explanation tonight. It's been a while and I forgot the exact details.

PR #171, opened 2018-02-14
support overriding record namespace per field

### LINKED ISSUE #170: Support overriding namespace at field level
Our use-case is that we have defined a common class that is used in multiple other classes. When we use `avro4s` to create the schemas for those classes the record definition for the common class is duplicated in each schema. We'd like to have a different namespace for that record definition in each schema that uses it. This is because the confluent schema registry doesn't support cross referencing schemas as highlighted [here](https://github.com/confluentinc/schema-registry/issues/426).

Currently if you put an `AvroNamespace` annotation at the case class field level it is just ignored. We will raise a PR shortly which will allow this.

We would like this to be the behaviour:

```scala
  case class Common(s: String, i: Int)

  case class Schema1(q: String, @AvroNamespace("hello") common: Common)

  case class Schema2(q: String, @AvroNamespace("olleh") common: Common)
```

The schemas this would generate are:

```json
{
	"type": "record",
	"name": "Schema1",
	"namespace": "com.sksamuel.avro4s",
	"fields": [{
		"name": "q",
		"type": "string"
	}, {
		"name": "common",
		"type": {
			"type": "record",
			"name": "Common",
			"namespace": "hello"
		}
	}]
}
```

```json
{
	"type": "record",
	"name": "Schema2",
	"namespace": "com.sksamuel.avro4s",
	"fields": [{
		"name": "q",
		"type": "string"
	}, {
		"name": "common",
		"type": {
			"type": "record",
			"name": "Common",
			"namespace": "olleh"
		}
	}]
}
```

lacarvalho91:
WIP: Closes #170

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/AvroSchemaTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/AvroSchemaTest.scala
@@ -315,6 +315,23 @@ class AvroSchemaTest extends WordSpec with Matchers {
       val schema = SchemaFor[AnnotatedNamespace]()
       schema.toString(true) shouldBe expected.toString(true)
     }
+
+    "support namespace annotations on field" in {
+      case class InternalAnnotated(i: Int)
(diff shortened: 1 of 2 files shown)


sksamuel:
:+1:

PR #188, opened 2018-04-09
Fix various Macro issues for 2.12

2chilled:


2chilled:
Hm strange, was pretty sure 2.11 would fail

sksamuel:
Nice job on this fix.

sksamuel:
It does fail for 2.11 for me locally.

PR #195, opened 2018-05-07
Making FromValue serializable (and therefore FromRecord).

wzorgdrager:
I made FromValue serializable which enables you to give FromRecord as a context bound but still have a serializable class.

As an example see this gist to use it with Apache Flink (and a KafkaConsumer/Producer):
https://gist.github.com/wzorgdrager/c27805ae9d627b57e736fe09dd09aad9


christophstockhusen:
Works like a charm!

sksamuel:
:+1:

Issue #200, opened 2018-05-21
org.apache.avro.SchemaParseException when multiple values of same generic user defined class are contained inside a case class


tmnd1991:
If I define two case classes like these:

```scala
case class Generic[T](a: T)
case class MyClass(a: Generic[Int], b: Generic[String])
```

And I try to generate an AvroOutputStream.data[Generic] like this:
```scala
val bos = new ByteArrayOutputStream()
val oos = AvroOutputStream.data[MyClass](bos).write(MyClass(Generic(3), Generic("")))    
```

I get a 
```org.apache.avro.SchemaParseException Can't redefine: Generic```

I think this is happening since the name of Generic[T] record type is always the classname both in the Int and String cases.

Probably this is fixable generating different type names for generic classes depending on the actual type(s).

Something like `Generic_Int` or `Generic_String` should work, right?

tmnd1991:
I opened PR #201

sksamuel:
Thanks for the PR. I think we'll have to release this as part of 2.0.0 release because this will break every single generated class that uses type parameters.

tmnd1991:
Thank you for the reply.
Do you have any plan on the 2.0.0 release?

sksamuel:
It shouldn't take longer than a month.

tmnd1991:
I was thinking about an annotation to switch on/off functionality but if you plan to release 2.0.0 in a month I don’t think it is worthwhile working on it. If you like the idea of making it optional I can work on it anyway!

sksamuel:
Let's do that, then I can release it as 1.9.1 immediately.

tmnd1991:
Wednesday and on friday I’ll have a 4 hour train connection I will give it a shot during that time and update the PR

tmnd1991:
I called the annotation `AvroSpecificGeneric`, I think it kinda fits ;)
I could not get rid of the `(enabled = true)` but does not look that bad.
If the implementation/interface looks nice to you I can write some docs in the README.md and then you can merge the PR to release 1.9.1.

tmnd1991:
We should close this I think ;)

PR #201, opened 2018-05-21
#200

tmnd1991:
suffix  type arguments name to generic classes in order to avoid type name clash


tmnd1991:
This is a PR for #200

sksamuel:
That was quick. There's a conflict so I can't merge. Also am I right in seeing that the default is for this feature to be off ?

tmnd1991:
I fixed the conflicts. Yes, the default is for the feature to be turned off. I can make it work the other way around if you think it is better. I think that leaving it off by default helps in case of an update without breaking "old" schemas. I would keep it disabled by default in 1.x.x and enable it by default on > 2.0.0.
What do you think?

sksamuel:
I agree that is the best approach.

PR #212, opened 2018-06-15
Support Scala Object Namespaces per default

2chilled:
Given something like 
```Scala
    case class Outer(inner: Outer.Inner)
    object Outer {
      final case class Inner(s: String)
    }
```
I think users expect the resulting Avro Schema to be something like:
 ```Javascript
{
  "type" : "record",
  "name" : "Outer",
  "namespace" : "com.sksamuel.avro4s",
  "fields" : [ {
    "name" : "inner",
    "type" : {
      "type" : "record",
      "namespace" : "com.sksamuel.avro4s.AvroSchemaTest.Outer",
      "name" : "Inner",
      "fields" : [ {
        "name" : "s",
        "type" : "string"
      } ]
    }
  } ]
}
```
Otherwise this would quickly lead to name clashes at Avro side while the Scala side remains fine. We ran into this while using a hand packaged version of avro4s which fully supports nested ADTs for Scala 2.12.

Of course, this patch is not very useful until the [patches](https://github.com/sksamuel/avro4s/compare/master...2chilled:avro_namespace_fix) for ADT support are merged into master. I suspect the requirement for this to happen is 2.11 support. Nevertheless, I think this improves Avro namespacing and prevents our fork from diverging too much.

### DIFF
--- a/avro4s-macros/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
+++ b/avro4s-macros/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
@@ -179,7 +179,7 @@ object FromValue extends LowPriorityFromValue {
     val from = implicitly[FromValue[T]]
 
     def typeName: String = {
-      val nearestPackage = Stream.iterate(tpe.typeSymbol.owner)(_.owner).dropWhile(!_.isPackage).head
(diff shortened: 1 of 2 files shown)


sksamuel:
This PR has been merged manually into the 2.0 branch. I've added another set of tests for it as well.

PR #213, opened 2018-06-18
Fixed broken default value accessor

2chilled:
The current accessor fails when the default value sits within a companion object. Atm this is just an issue on my master, as the avro4s master does not support ADTs with `object`-namespacing at all. But I want to keep sync as far as possible.


sksamuel:
:+1:

PR #216, opened 2018-06-19
Fix #215: make ToRecord macro obey AvroName annotation

### LINKED ISSUE #215: Serializing / Deserializing not working when using an AvroName annotation
The basic functionality for generating a schema with a different field name based on the AvroName annotation (as described in Issue #159) is working and the generated schema looks ok. 

However, the AvroName annotation is not respected when serializing/deserializing and generates an exception:

> org.apache.avro.AvroRuntimeException: Not a valid schema field: fat

Snippet based on the official samples showing the behaviour:
```
import java.io.File
import com.sksamuel.avro4s.{AvroName, AvroSchema, AvroOutputStream}

case class Ingredient(name: String, sugar: Double, @AvroName("fat_renamed") fat: Double)

case class Pizza(name: String, ingredients: Seq[Ingredient], vegetarian: Boolean, vegan: Boolean, calories: Int)

val schema = AvroSchema[Pizza]
schema.toString(true)

// Serialize
val pepperoni = Pizza("pepperoni", Seq(Ingredient("pepperoni", 12, 4.4), Ingredient("onions", 1, 0.4)), false, false, 98)
val hawaiian = Pizza("hawaiian", Seq(Ingredient("ham", 1.5, 5.6), Ingredient("pineapple", 5.2, 0.2)), false, false, 91)

val os = AvroOutputStream.data[Pizza](new File("pizzas.avro"))
os.write(Seq(pepperoni, hawaiian))
os.flush()
os.close()
```

felixmulder:
This PR fixes the serialization issues we had with `@AvroName`

@sksamuel -- would it be possible for you to make a release with this fix once merged?

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/AvroNameTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/AvroNameTest.scala
@@ -13,4 +13,11 @@ class AvroNameTest extends WordSpec with Matchers {
       schema.toString(true) shouldBe expected.toString(true)
     }
   }
+
+  "ToRecord" should {
+    "correctly be able to produce a record" in {
+      val toRecord = ToRecord[Foo]
+      toRecord(Foo("woop", "scoop"))
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


felixmulder:
Turns out that `FromRecord` had to be fixed as well.

Issue #223, opened 2018-09-08
FromValue / ToValue serializability


ckchow:
Is it feasible to make FromValue and ToValue serializable?

Zhen-hao:
I encountered the same problem. 
It seems that making ToRecord and FromRecord serializable is not enough. Because the SchemaFor macro can pull in non-serializable objects. 
For that reason, I find it very hard to use this library in Spark and Flink.  

It also relates to https://github.com/sksamuel/avro4s/issues/19#issuecomment-385678804

sksamuel:
In the 2.0 development branch the equivalent classes (Encoder and Decoder, might be renamed yet) are serializable.

sksamuel:
If you want to try 2.0.0-M3, all the required traits are now serializable.

sksamuel:
2.0.0 is now released, so closing this ticket.

Issue #19, opened 2016-04-26
How to use Avro4s to create generic KafkaDecoder for Spark streaming


vladap:
I would like to use your great work but I'm stuck with my limited knowledge of Scala to solve the following problem. I know you possible don't know Kafka and Spark but it is more about reflection and context bounds.

RecordFormat[T] requires fromRecord and toRecord implicit values. Spark KafkaDirectStream takes Kafka Decoder type and creates instance of it via reflection looking for a single argument constructor. When fromRecord and toRecord are provided in decoder implementation via context bounds or as implicit values in separate parameter list compiler add them to constructor as additional arguments breaking the required contract.

Does anybody know how to provide these implicits in KafkaAvroGenericDecoder shown below so that RecordFormat is satisfied without breaking Spark code?

Goal is to let avro4s generate decoding from Avro GenericRecord to case class using Confluent Schema Registry.

Spark code to satisfy, VD is Decoder type and V is resulting case class, Decoder is interface defined by Kafka:
```scala
val valueDecoder = classTag[VD].runtimeClass.getConstructor(classOf[VerifiableProperties])
  .newInstance(props)
  .asInstanceOf[Decoder[V]]
```

The code I currently have, context bounds removed:

```scala
class KafkaAvroGenericDecoder[T]
  extends AbstractKafkaAvroDeserializer
  with Decoder[T] {

  def this(schemaRegistry: SchemaRegistryClient) = {
    this()
    this.schemaRegistry = schemaRegistry
  }

  def this(schemaRegistry: SchemaRegistryClient, props: VerifiableProperties) = {
    this(schemaRegistry)
    configure(deserializerConfig(props))
  }

  /** Constructor used by Kafka consumer */
  def this(props: VerifiableProperties) = {
    this()
    configure(new KafkaAvroDeserializerConfig(props.props))
  }

  override def fromBytes(bytes: Array[Byte]): T = {
    val record = deserialize(bytes).asInstanceOf[GenericRecord]
    val format = RecordFormat[T]
    format.from(record)
  }
}

object Kafka {
  def createStream[V : ClassTag](topics: Set[String])
    (implicit ssc: StreamingContext): InputDStream[(Array[Byte], V)] = {
  
    KafkaUtils.createDirectStream[Array[Byte], V, DefaultDecoder, KafkaAvroGenericDecoder[V]](
      ssc, session.kafkaConf, topics)

    // Or this one for test. Change return type to InputDStream[Array[Byte]]
[...]

PR #224, opened 2018-09-17
ToValue and LowPriorityToValue extends Serializable

Zhen-hao:
I think this will fix the issues mentioned in https://github.com/sksamuel/avro4s/issues/19#issuecomment-385678804 and https://github.com/sksamuel/avro4s/issues/223


sksamuel:
:+1:

PR #230, opened 2018-10-09
Respect avroname in sealed case object enums

alexjg:
I noticed that `@AvroName` is not respected in enum branches for `case object` style enums. I've added a few test cases and got them passing but I don't know if the solution I've come up with is the best one. Please let me know if there are changes I should make.

### DIFF
--- a/avro4s-macros/src/main/scala/com/sksamuel/avro4s/Decoder.scala
+++ b/avro4s-macros/src/main/scala/com/sksamuel/avro4s/Decoder.scala
@@ -299,7 +299,7 @@ object Decoder extends CoproductDecoders with TupleDecoders {
     private def getTypeName[T](clazz: Class[T]): String = {
       val mirror = runtimeMirror(clazz.getClassLoader)
       val tpe = mirror.classSymbol(clazz).toType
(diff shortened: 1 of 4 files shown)


avro4s-macros/src/main/scala/com/sksamuel/avro4s/Decoder.scala
```diff
@@ -289,8 +289,19 @@ object Decoder extends CoproductDecoders with TupleDecoders {
                                                                      toList: ToList[L, T]): Decoder[T] = new Decoder[T] {
     override def decode(value: Any, schema: Schema): T = {
       val name = value.toString
-      toList(objs()).find(_.toString == name).getOrElse(sys.error(s"Uknown type $name"))
+      val variants = toList(objs())
+      variants.find(v => {
+        println(getTypeName(v.getClass))
```
sksamuel: Should remove this.

sksamuel:
I think it looks great bar leaving in the println.

alexjg:
Whoops, removed!

sksamuel:
Excellent

PR #231, opened 2018-10-10
Decode UUIDs from org.apache.avro.utl.Utf8

alexjg:
Another small PR. Noticed that there's no branch to handle `Utf8` when decoding `UUID`s so added a test and implementation.


sksamuel:
Good spot.

Issue #235, opened 2018-10-11
Broken typeclass derivation upgrading from 1.9.0 to 2.0.1


Astrac:
I am using avro4s to auto-generate kafka serdes with a function that looked like this in 1.9.0:

```scala
object AutoSerde {
  def apply[A](
      implicit fromRecord: FromRecord[A],
      toRecord: ToRecord[A],
      schemaFor: SchemaFor[A]
  ): Serde[A] = {

    val schema = schemaFor()
    val datumReader = new GenericDatumReader[GenericRecord](schema)
    val datumWriter = new GenericDatumWriter[GenericRecord](schema)
    ... // Uses datumReader and datumWriter to implement the serde
  }
}
```

Upgrading to 2.0.1 it looks like this:

```scala
object AutoSerde {
  def apply[A <: Product](
      implicit schemaFor: SchemaFor[A],
      encoder: Encoder[A],
      decoder: Decoder[A]
  ): Serde[A] = {

    val format = RecordFormat[A]
    val schema = schemaFor.schema
    val datumReader = new GenericDatumReader[GenericRecord](schema)
    val datumWriter = new GenericDatumWriter[GenericRecord](schema)
    ... // Uses datumReader and datumWriter to implement the serde
  }
}
```

I also have a property-based test that serialises and deserialises this ADT usign a `Serde` generated this way:

```scala
  object AutoSerdeSpecModel {
    case class Label(value: String) extends AnyVal
    case class Value[A](label: Label, value: A)

    sealed trait OneOrTwo[A]
    case class One[A](value: Value[A]) extends OneOrTwo[A]
    case class Two[A](first: Value[A], second: Value[A]) extends OneOrTwo[A]

    case class OneOrTwoWrapper[A](t: OneOrTwo[A])
  }

  class MySpec extends ... {
    def checkOne[T: Arbitrary](serde: Serde[T])(beforeSerde: T): Boolean = {
      val afterSerde = serde
        .deserializer()
        .deserialize(topic, serde.serializer().serialize(topic, beforeSerde))

      afterSerde == beforeSerde
    }

    "Derive a working serde for a generic coproduct type" in {
      Checkers.check(property(AutoSerde[OneOrTwoWrapper[Int]]))
    }
  }
```

In the 1.9.0 version everything works fine, in the 2.0.1 it fails with the error:

```
 Cannot find subschema for class my.package.AutoSerdeSpecModel.One
```

or:

```
[...]

PR #236, opened 2018-10-11
WIP: Fix Inconsistency in namespace calculation between `Encoder` and `SchemaFor`

alexjg:
There appears to be an inconsistency in the way that namespaces are calculated for case classes that are defined inside an object between `SchemaFor[T]` and `Encoder[T]`. This merge request contains a test case that reproduces it. Specifically the inconsistency is below:

`Encoder.applyMacroImpl`
```scala
  def applyMacroImpl[T: c.WeakTypeTag](c: scala.reflect.macros.whitebox.Context): c.Expr[Encoder[T]] = {

    import c.universe._

    val reflect = ReflectHelper(c)
    val tpe = weakTypeTag[T].tpe
    val fullName = tpe.typeSymbol.fullName
    ...
          } else {
        c.Expr[Encoder[T]](
          q"""
            new _root_.com.sksamuel.avro4s.Encoder[$tpe] {
              override def encode(t: $tpe, schema: org.apache.avro.Schema): AnyRef = {
                _root_.com.sksamuel.avro4s.Encoder.buildRecord(schema, Seq(..$fields), $fullName)
              }
            }
        """
        )
```

And 

`SchemaFor.applyMacroImpl`
```scala
  def applyMacroImpl[T: c.WeakTypeTag](c: whitebox.Context): c.Expr[SchemaFor[T]] = {
    import c.universe
    import c.universe._

    val reflect = ReflectHelper(c)
    val tpe = weakTypeOf[T]
    val fullName = tpe.typeSymbol.fullName
    val packageName = reflect.packageName(tpe)
    ...
          // if the default getter exists then we can use it to generate the default value
          if (defaultGetterMethod.isMethod) {
            val moduleSym = tpe.typeSymbol.companion
            q"""{ _root_.com.sksamuel.avro4s.SchemaFor.schemaFieldWithDefault[$fieldTpe]($fieldName, $packageName, Seq(..$annos), $moduleSym.$defaultGetterMethod) }"""
          } else {
            q"""{ _root_.com.sksamuel.avro4s.SchemaFor.schemaFieldNoDefault[$fieldTpe]($fieldName, $packageName, Seq(..$annos)) }"""
          }
        }
```

You can see that in `Encoder` the `TermSymbol.fullName` is used whereas in `SchemaFor` `ReflectHelper.packageName` is used. `ReflectHelper.packageName` returns (unsurprisingly) the package name - i.e not including the surrounding object that a case class nested in an object contains in it's full name.

This means that for example the following code

```
package mypackage 

object Outer {
    case class MyRecord(value: Double)
    val schema = SchemaFor[MyRecord]
}
```

would produce a schema with the namespace of `MyRecord` being set to `mypackage`, but when attempting to encode `Encoder[MyRecord]` would be looking for a record with namespace `mypackage.Outer`. (This only manifests in sealed trait unions I think because otherwise there is no reason for `Encoder` to look up the record).

I've included a possible fix in this merge request. There are a lot of failing tests because this change changes a lot of namespaces. Before I went ahead and fixed all those tests I thought it would be worth asking whether it's actually desirable to have the namespace include the surrounding object? Would it be better to issue a warning of some kind?

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/encoder/SealedTraitEncoderTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/encoder/SealedTraitEncoderTest.scala
@@ -54,6 +54,18 @@ class SealedTraitEncoderTest extends FunSuite with Matchers {
     Encoder[Dibble].encode(Dobble, schema) shouldBe new GenericData.EnumSymbol(schema, Dobble)
     Encoder[Dibble].encode(Dabble, schema) shouldBe new GenericData.EnumSymbol(schema, Dabble)
   }
+
+  test("classes nested in objects should be encoded with consistent package name") {
+    sealed trait Inner
+    case class InnerOne(value: Double) extends Inner
+    case class InnerTwo(height: Double) extends Inner
+    case class Outer(inner: Inner)
+    val schema = AvroSchema[Outer]
+    val record = Encoder[Outer].encode(Outer(InnerOne(1.23)), schema).asInstanceOf[GenericRecord]
+    val inner = record.get("inner").asInstanceOf[GenericRecord]
+    inner.get("value") shouldBe 1.23
+  }
+
 }
 
 sealed trait Dibble
--- a/avro4s-macros/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
+++ b/avro4s-macros/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
@@ -255,6 +255,7 @@ object SchemaFor extends TupleSchemaFor with CoproductSchemaFor {
 
       val isValueClass = reflect.isValueClass(tpe)
       val recordName = reflect.recordName(tpe)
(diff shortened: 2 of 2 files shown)


sksamuel:
So the rationale, and original PR, was here:
https://github.com/sksamuel/avro4s/pull/212

I think the reason is if you have classes inside different objects, then they should be namespaced independently.

So

object Foo {
  case class Sam()
}

and

object Boo {
  case class Sam()
}

Two completely different Sams.

sksamuel:
Also, thanks for the PR. There is a conflict atm.

alexjg:
Okay great, I'll go ahead and fix conflicts and implement. It might make sense in the same piece of work to implement using the `NamingStrategy` and `AvroName` stuff in `SchemaHelper.extractTraitSubschema` as it's causing similar problems. I'll have a bash at that too.

sksamuel:
Awesome, be nice to cover any gaps.

sksamuel:
I've made some changes for #235 so that AvroNameResolver now has all the methods needed to correct resolve the name and namespaces for a type. This should prove useful in this pr.

PR #238, opened 2018-10-11
Fix an error where options of sealed traits weren't encoded correctly

alexjg:


sksamuel:
I assume this is a fix is because the coproduct encoder requires the union schema recursively?

PR #239, opened 2018-10-11
Improving name generation in encoder

alexjg:
This is a slightly improved approach to respecting `@AvroNamespace` in `Encoder`, it closes #236 . This MR also implements using outer classes when generating namespaces, which is why there are so many changes. I've kept all the mechanical changes to the tests in one commit (Fix ALLLL the tests) so probably easiest to review by looking at the other two commits.

One thing to note is that this still doesn't use the `NamingStrategy` in the `Encoder`, although I think this should be pretty easy to tack on as an implicit argument to the new `NameResolution` class.


sksamuel:
I've only just seen this and my changes have caused conflicts. I'll do the conflict resolution. Does this PR replace the earlier one ?

PR #240, opened 2018-10-12
Respect AvroName when generating and encoding enum symbols 

alexjg:
Respect AvroName when generating and encoding enum symbols for sealed trait case object hierarchies


sksamuel:
Excellent.

Issue #234, opened 2018-10-10
RecordFormat macro for List - diverging implicit expansion for type 


vilinski:
avro4s 2.0.1
scala 2.12.7

the following code doesn't work for me

```scala
import com.sksamuel.avro4s._
object Formats {
    case class TestEntry(name: String)
    case class Test(id: Int, entries: List[TestEntry])
    val test: RecordFormat[Test] = RecordFormat[Test] // <---
}
```

however if `entries` is a `List[Int]` it compiles. Do I something wrong?

sksamuel:
I added a test with the classes you've pasted here and it works ok for me.
https://github.com/sksamuel/avro4s/blob/master/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue234.scala

Can you make a test that shows this - perhaps add one to this code base and PR it back as a failing test?

sksamuel:
-XLog-implicits is a bit hard to understand for sure.

The only issue I know of at the moment is a sequence of maps of case classes. Are you mixing maps in your real code rather than the example?

vilinski:
I have no `Map`s from which I'm aware. One of case classes has `BsonValue` field, and `BsonDocument` implements `java.util.Map<String, BsonValue>`. But it's not a class which brings compile errors and where I have custom serializer:

```scala
  implicit object BsonValueToSchema extends SchemaFor[BsonValue] {
    override val schema: Schema = Schema.create(Schema.Type.STRING)
  }
  implicit object BsonValueFromValue extends Decoder[BsonValue] {
    override def decode(value: Any, schema: Schema): BsonValue = value match {
      case s: CharSequence => BsonDocument(s.toString).get("value")
      case other           => sys.error("Unsupported type for BsonValue " + other)
    }
  }
  implicit object BsonValueToValue extends Encoder[BsonValue] {
    override def encode(value: BsonValue, schema: Schema): String = Document("value" -> value).toJson()
  }
```

I have also some ADT where each case gets `RecordFormat` compiled, but a case class containing a trait not. Don't get why. 
I would also mention, anything works with 1.8.3, but not with 2.0.1. Is there anywhere a changelog?
By the way, could it be possible to support sealed class as ADTs additionally to traits?

sksamuel:
If shapeless supports coproducts on sealed classes then yes, but I don't know if it does. Just try it in avro4s, if it works then you're good.

[...]

PR #244, opened 2018-10-18
Diverging implicit expansion for sealed case class

alexjg:
I can't figure out what's causing this so haven't got a fix yet, but this is a testcase for something that's biting me at the moment. Causes errors like this:

```
[error] <base dir>/avro4s/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/encoder/SealedTraitEncoderTest.scala:70: diverging implicit expansion for type com.sksamuel.avro4s.SchemaFor[Outer]
[error] starting with macro method applyMacro in object SchemaFor
[error]     val schema = AvroSchema[Outer]
```

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/SealedTraitDecoderTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/SealedTraitDecoderTest.scala
@@ -75,6 +75,10 @@ class SealedTraitDecoderTest extends FunSuite with Matchers {
     Decoder[ThingHolder].decode(record1, schema) shouldBe ThingHolder(WhimWham)
     Decoder[ThingHolder].decode(record2, schema) shouldBe ThingHolder(Widget)
   }
+
+  test("sealed trait classes with optional elements inside case classes should derive withou implicit divergence") {
+    val d = Decoder[Coat]
+  }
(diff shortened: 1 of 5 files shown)


sksamuel:
It will be related to #234

alexjg:
So I've got a fix for the divergence problems but it's causing these weird `ClassNotFoundException`s in the tests.

coltfred:
@alexjg Are you still working on this? I ran into this issue as well and eagerly await a fix so I can move to 2.x. Thank you to both you and @sksamuel for your work!

sksamuel:
I've added your test case and got it working in 2.0.3
Can you try that version ?

coltfred:
@sksamuel In my initial compilation, 2.0.3 seems to work for my use case. I'll spend some time today trying to do the full upgrade to the 2.x series and see if everything works well.

sksamuel:
This should be working on master now (grab the latest snapshot).

gregsilin:
@alexjg thanks for your work

@sksamuel is this a fix for https://github.com/sksamuel/avro4s/issues/254?  I may be getting things confused here.

sksamuel:
254 is fixed in 3.0.x. You can test the latest snapshot by clicking on the badge on the homepage.

gregsilin:
Great, thanks!

PR #246, opened 2018-10-19
Issue-242

ljurukov:
Added shadow traits for the case classes so that the annotations are extensible

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue193.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue193.scala
@@ -2,11 +2,10 @@ package com.sksamuel.avro4s.github
 
 import java.io.ByteArrayOutputStream
 
-import com.sksamuel.avro4s._
+import com.sksamuel.avro4s.{AvroFixed, AvroInputStream, AvroOutputStream, AvroSchema}
 import org.scalatest.{FunSuite, Matchers}
 
-@AvroName("Data")
-case class Data(@AvroName("name")uuid: Option[UUID])
+case class Data(uuid: Option[UUID])
 case class UUID(@AvroFixed(8) bytes: Array[Byte])
 
 class GithubIssue193 extends FunSuite with Matchers {
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue193.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue193.scala
@@ -2,10 +2,11 @@ package com.sksamuel.avro4s.github
 
 import java.io.ByteArrayOutputStream
 
-import com.sksamuel.avro4s.{AvroFixed, AvroInputStream, AvroOutputStream, AvroSchema}
+import com.sksamuel.avro4s._
 import org.scalatest.{FunSuite, Matchers}
 
-case class Data(uuid: Option[UUID])
+@AvroName("Data")
+case class Data(@AvroName("name")uuid: Option[UUID])
 case class UUID(@AvroFixed(8) bytes: Array[Byte])
 
 class GithubIssue193 extends FunSuite with Matchers {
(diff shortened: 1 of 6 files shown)


sksamuel:
Looks good, the other thing I thought, is do the annotations have to be case classes, because with regular classes you could just extend them?

ljurukov:
For one, new annotation, to many, base annotation, extensibility still be limited if they were just classes

For one to one extensibility it might make sense to make them as classes rather then case classes.

Regardless the annotations being case classes doesn't really result in a difference other then the removal of "override val" in front of every variable.

You question did prompt another thought that I just tested out.

With the one to many trait methodology @AvroProperty will need further changes in two places.
The first being that there is a variable collision now between it and @AvroName to which I propose changing the name from "name" -> "key"

The other, hairier, issue is with the reflection code that pulls out the arguments in the `case class Anno(className: String, args: Seq[Any])` seems to be having issues when the @AvroProperty is mixed in with other properties. Example, using your cold play test:

`class AvroProp2(override val value: String) extends AvroNameable with AvroProperty {
  override final val key: String = "cold"
  override val name: String = "static"
}`

results in: 

> {
  "type" : "record",
  "name" : "play",
  "namespace" : "com.sksamuel.avro4s.schema.AvroPropSchemaTest",
  "fields" : [ {
    "name" : "str",
    "type" : "string"
  } ],
  "play" : "avro4s.this.AvroProp2.<init>$default$2"
}

When @AvroProp2("play") is used

I can either close the PR until this is figured out. Or make them all the turn all "case class" -> "class" and turn the "traits" into "sealed traits" until other code is figured out in a separate PR

ljurukov:
@sksamuel - Follow up to the things I found earlier and a proposed fix for it.
Now instead of relying on an implicit understanding of the structure of the case classes to pull out the fields the code now pulls out the fields as a Map, with the field name being the key.

This was done by compiling the class and using reflection to pull out all the relevant fields. This was done instead of the methodology before due to the fact that it only gave the fields, and only gave those fields that were defined in the constructor.

This methodology does come with some caveats though. It seems to only work if the annotations are case classes (which they were to being with), not exactly sure why. And doesn't work if there are default parameters within the structure (this is due to the fact that the toolbox used to compile the code isn't able to handle relative package names, which is what defaults are). Fortunately this was not a previously supported feature either.

This now allows the creation of custom annotations that allow the user to define composite annotations.

Let me know what you think, I'm still pretty new to macros so if there are better ways to do what I accomplished please let me know.

Final note: Strangely writing a unit test with a custom annotation failed to compile as long as the annotation was in the test package scope, if it is not it works just as expected. I'm not sure as to the reason for this, so I'm not sure the best way to write the unit test for custom annotations

sksamuel:
Does this rely on the values in each annotation to have different names ?

sksamuel:
Ok that's what I thought. It's not ideal that every annotation that might want to be "mixed in" has to have a different name for it's attribute(s). Name is gone, key is gone, if we want to add another for something, we need to come up with another.

ljurukov:
I'm not sure how exactly this would be avoided. While not ideal, annotations currently map to specific keys in the avro spec so overlap shouldn't be a huge issue if the variable names match those key names. This could still be an issue is in the case of overlapping annotations (two annotations covering the same Avro key name), in which case the "mixed in" functionality should cover this. Perhaps in some instances more specific names can be used? like "key" -> "propertyKey"

ljurukov:
Gave it a little more thought, it seems like for the case of "key" once this is merged I can work on the `AvroProps` annotation (Issue 243) which would have a `properties` value of type Map[String,String]. `AvroProp` can then have the current signature but place it's values in the `properties` map internally and inheriting the AvroProps shadow trait. This should free `key` and `value` for future use.

sksamuel:
I guess you're right in that every annotation at the class level maps to different fields in avro so as long as our annotation names == the avro field name, voila.

Plus since they haven't done a release in 18 months, I don't see Avro adding too many new ones anytime soon.

PR #248, opened 2018-10-28
fix Option TraitObjectEnum not serializing

coreyauger:
Hey there !
This might be a bit of a long shot.. 
When the fix for `Option[ Enum ]` made it in to version `1.8.3`, it was not applied to `TraitObjectEnum`.
I have a project that makes extensive use of trait object enums and is not ready to migrate to 1.9+

Is it possible to sneak this PR in as a minor release fix?  
I realize that you are probably no longer supporting this version.. 
but you can always ask :)

Thanks!


sksamuel:
This has a conflict because master is on v2.0.2
If you do the PR against the release/1.8 branch I will merge and release 1.8.4

PR #258, opened 2019-01-03
Improve decode performance of coproducts and eithers

jeroentervoorde:
Performance improvements for coproducts and either types:
- Speed up safeFrom by reusing the Type instance and avoiding runtime typechecks on the branch type.
- Avoid recreating Decoder instances in the macro.

I tried creating vals for each decoder in the macro to avoid the ugly typecasting but i couldn't get that to work.

I had to move the Class.forName in the enum decoder to the decode method to avoid a classloading issue. I suspect this used to work because the Decoders were recreated at runtime.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/FromRecord.scala
@@ -13,6 +13,7 @@ trait FromRecord[T] extends Serializable {
 object FromRecord {
   def apply[T: Decoder : SchemaFor]: FromRecord[T] = apply(AvroSchema[T])
   def apply[T: Decoder](schema: Schema): FromRecord[T] = new FromRecord[T] {
-    override def from(record: GenericRecord): T = implicitly[Decoder[T]].decode(record, record.getSchema)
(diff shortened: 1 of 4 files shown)


jeroentervoorde:
> > I tried creating vals for each decoder in the macro to avoid the ugly typecasting but i couldn't get that to work.
> 
> Are you talking about this bit:
> 
> ```
> val decoders = reflect.constructorParameters(tpe).zipWithIndex.map { case ((fieldSym, fieldTpe), index) =>
> 
>           if (reflect.isMacroGenerated(fieldTpe)) {
>             q"""implicitly[_root_.com.sksamuel.avro4s.Decoder[$fieldTpe]]"""
>           } else {
>             q"""implicitly[_root_.shapeless.Lazy[_root_.com.sksamuel.avro4s.Decoder[$fieldTpe]]]"""
>           }
>         }
> ```

Yes. I tried to generate something like 'val decoder_$idx: Decoder[$fieldType] = implicitly[...] for each field and just passing that value in calls to decodeFieldOrApplyDefault[Not]Lazy invocations below that but i couldn't get the macro to work.

Anyway, thanks for merging.

PR #269, opened 2019-01-14
Issue-268: Don't check SchemaName/Namespace when extracting subschema from Union if subschema is obvious

zzeekk:


sksamuel:
:+1:

Issue #267, opened 2019-01-10
BigDecimal scale and precision when suppling a schema


sebastianvoss:
I'm using the Confuent schema registry to store my schema and running into issues when reading records with `BigDecimal` values. I would like to use the schema supplied by the registry rather than letting `avro4s` generate it. But when I supply a schema to `RecordFormat[Measurement](schema)` it seems it is not repecting `scale` and `precision` for logical type `decimal`.

Here is an example:

```
case class Measurement(value: BigDecimal)

val deserializer = new GenericAvroDeserializer()
val record = deserializer.deserialize("test", bytes)
val schema = record.getSchema
val format = RecordFormat[Measurement](schema)
val m = format.from(record)
```
When I'm running the same code with `implicit val sp = ScalePrecisionRoundingMode(8, 20, BigDecimal.RoundingMode.HALF_UP)` in scope it gives me different results (e.g. `123456789.12312313` with `sp` in scope  and `123456789123123.13` without).

In the registry, the schema for field `value` is:
```
{
    "name" : "value",
    "type" : {
      "type" : "bytes",
      "logicalType" : "decimal",
      "precision" : 20,
      "scale" : 8
    }
}
```

Shouldn't `avro4s` take the precision and scale from the supplied schema and not requiring an implicit `ScalePrecisionRoundingMode` in this case?

sksamuel:
Yes that's right. The Decoder should use the values in the schema.
Only the schema generator should use `ScalePrecisionRoundingMode`.

sksamuel:
Fixed in 2.0.4

PR #276, opened 2019-01-27
Fix bigdecimal encoder and decoder

olib963:
Work in progress, happy to take feedback. In the process of fixing the bigdecimal en/decoders I uncovered a couple of issues in the decoders for `String`, `Option` and most collections. I hope the fixes are acceptable.

I am also looking at allowing `BigDecimal` -> `FIXED` in the decoder & encoder unless you feel this should be a separate PR.

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/FixedDecoderTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/FixedDecoderTest.scala
@@ -20,7 +20,7 @@ class FixedDecoderTest extends FunSuite with Matchers {
   test("support options of fixed") {
     val schema = AvroSchema[OptionalFixedValueType]
     val record = new GenericData.Record(schema)
-    record.put("z", Array[Byte](115, 97, 109))
+    record.put("z", new GenericData.Fixed(AvroSchema[FixedValueType], Array[Byte](115, 97, 109)))
     Decoder[OptionalFixedValueType].decode(record, schema) shouldBe OptionalFixedValueType(Some(FixedValueType("sam")))
   }
 }
--- a/avro4s-macros/src/main/scala/com/sksamuel/avro4s/Decoder.scala
+++ b/avro4s-macros/src/main/scala/com/sksamuel/avro4s/Decoder.scala
@@ -189,56 +189,28 @@ object Decoder extends CoproductDecoders with TupleDecoders {
     override def decode(value: Any, schema: Schema): Option[T] = if (value == null) None else {
       // Options are Union schemas of ["null", other], the decoder may require the other schema
       val nonNullSchema = schema.getTypes.asScala.filter(_.getType != Schema.Type.NULL).toList match {
-        case schema :: Nil => schema
+        case s :: Nil => s
         case multipleSchemas => Schema.createUnion(multipleSchemas.asJava)
       }
       Option(decoder.decode(value, nonNullSchema))
     }
   }
 
-  implicit def vectorDecoder[T](implicit decoder: Decoder[T]): Decoder[Vector[T]] = new Decoder[Vector[T]] {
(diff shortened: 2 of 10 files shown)


olib963:
I believe this should address https://github.com/sksamuel/avro4s/issues/267.

@sksamuel I was also wondering if instead of `ScalePrecisionAndRoundingMode` it would be better to use an annotation on BigDecimal fields? e.g.

```
case class Foo(@AvroScalePrecision(4, 10) bar: BigDecimal, // scale = 4, precision = 10
                          baz: BigDecimal) // scale = 2, precision = 8 as default

File: avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/FixedDecoderTest.scala [hunk 1]
```diff
@@ -20,7 +20,7 @@ class FixedDecoderTest extends FunSuite with Matchers {
   test("support options of fixed") {
     val schema = AvroSchema[OptionalFixedValueType]
     val record = new GenericData.Record(schema)
-    record.put("z", Array[Byte](115, 97, 109))
+    record.put("z", new GenericData.Fixed(AvroSchema[FixedValueType], Array[Byte](115, 97, 109)))
```
sksamuel: Should we keep both tests since Fixed doesn't always have to be GenericData.Fixed ?

avro4s-macros/src/main/scala/com/sksamuel/avro4s/Decoder.scala
```diff
@@ -184,7 +186,14 @@ object Decoder extends CoproductDecoders with TupleDecoders {
...
-    override def decode(value: Any, schema: Schema): Option[T] = if (value == null) None else Option(decoder.decode(value, schema))
+    override def decode(value: Any, schema: Schema): Option[T] = if (value == null) None else {
+      // Options are Union schemas of ["null", other], the decoder may require the other schema
+      val nonNullSchema = schema.getTypes.asScala.filter(_.getType != Schema.Type.NULL).toList match {
+        case schema :: Nil => schema
+        case multipleSchemas => Schema.createUnion(multipleSchemas.asJava)
+      }
+      Option(decoder.decode(value, nonNullSchema))
+    }
   }
```
sksamuel: :+1:

avro4s-macros/src/main/scala/com/sksamuel/avro4s/Decoder.scala
```diff
@@ -244,15 +253,11 @@ object Decoder extends CoproductDecoders with TupleDecoders {
...
-        val decimalType = LogicalTypes.decimal(sp.precision, sp.scale)
-        val bytes = value.asInstanceOf[ByteBuffer]
-        decimalConversion.fromBytes(bytes, null, decimalType)
-      }
-    }
+  implicit object BigDecimalDecoder extends Decoder[BigDecimal] {
+    private val decimalConversion = new Conversions.DecimalConversion
+    override def decode(value: Any, schema: Schema): BigDecimal =
+      // TODO if the encoder is allowing encoding to strings should this not also allow String decoding?
+      decimalConversion.fromBytes(value.asInstanceOf[ByteBuffer], schema, schema.getLogicalType)
```
sksamuel: :+1:

avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/FixedDecoderTest.scala
olib963: It doesn't? I didn't realise that :P. I will go and change that if you like so that the decoder can handle both?

sksamuel: Yeah because people are allowed to make records themselves, and a normal array is fine (although GenericData.Fixed is what Avro itself would use when marshalling).

PR #282, opened 2019-01-29
Fix the output of the AvroDoc example in the readme.

philip-ovo:
This fixes a couple of errors in the schema given as the output of the `@AvroDoc` example in the readme.


sksamuel:
Thanks @philip-ovo

Issue #284, opened 2019-01-31
RecordFormat[T].from() cannot take SpecificRecord (v2.0.4)


ChengGit:
Hi,

Based on the documentation on `Avro Records` v2.0.4 (see link: https://github.com/sksamuel/avro4s/blob/master/README.md#avro-records)

My understanding is that now avro4s support bi-directional conversion between a case class `T` and `avro4s.Record` by doing the following:
 - T to Record:   RecordFormat[T].to(`T instance`)
 - Record to T: RecordFormat[T].from(`Record instance`)

However, in the current version, the `RecordFormat[T].from` still takes a `GenericRecord` instead of a `Record`( see link: 
 https://github.com/sksamuel/avro4s/blob/4f8a4f14a94f209bb3685a80370d54d5ae37892a/avro4s-core/src/main/scala/com/sksamuel/avro4s/FromRecord.scala#L10)

May I know if the type of parameter `RecordFormat[T].from` is supposed to be a `Record` instead?

I am asking because we are current de/serialise our model with `io.confluent.kafka.streams.serdes.avro.SpecificAvroSerde` to convert between specific record and binary code. So I would love to see that we could utilise `avro4s`'s power to convert between specific record and case class.

thanks for your time!

sksamuel:
I think it might be able to have `FromRecord` accept a `IndexedRecord` which is the parent of both.

ChengGit:
hey @sksamuel , thanks for the suggestion! now I have some time to work on it and I drafted this pr to address the issue aforementioned, if you would like to have a look?
https://github.com/sksamuel/avro4s/pull/296
Thanks for your time!

bszmit:
Could you publish a new version with this change?

sksamuel:
Can you try the 3.0.x snapshot from 
https://oss.sonatype.org/content/repositories/snapshots/com/sksamuel/avro4s
This should be fixed and is available for 2.11 and 2.22

bszmit:
It works.
Thanks!

PR #296, opened 2019-03-11
RecordFormat.from accepts IndexRecord

ChengGit:
Applied suggested change by update the parameter type from GenericRecord to IndexedRecord, in order to handle conversion between a specific record and case class


ChengGit:
this pr is related to issue #284 
may I have your opinion it please?

PR #301, opened 2019-04-11
Fix SchemaFor when either have nested union types

klass-ivan:
```
AvroSchema[Either[String, Option[Int]]]
org.apache.avro.AvroRuntimeException: Nested union: ["string",["null","int"]]
  at org.apache.avro.Schema$UnionSchema.<init>(Schema.java:849)
  at org.apache.avro.Schema.createUnion(Schema.java:189)
  at org.apache.avro.Schema.createUnion(Schema.java:194)
  at com.sksamuel.avro4s.SchemaFor$$anon$5.schema(SchemaFor.scala:92)
  at com.sksamuel.avro4s.AvroSchema$.apply(AvroSchema.scala:21)
```


klass-ivan:
This fixes a schema, but does not prevent decode errors :/ The decoder just tries `schema.getTypes(0)` and `schema.getTypes(1)` which is not correct with nested unions. Something like an approach of flattened coproduct should be used

Issue #273, opened 2019-01-22
Diverging implicit expansion for SchemaFor in Coproducts inside case classes


fedefernandez:
Hi @sksamuel, just found this while upgrading one project from `1.8.3` to `2.0.2` but seems its also reproducible in master.

Given this model:
```scala
case class Coproducts(cp: Int :+: String :+: Boolean :+: CNil)
case class CoproductOfCoproductsField(cp: Coproducts :+: Boolean :+: CNil)
```

Trying to get a `SchemaFor[CoproductOfCoproductsField]` gives you a diverging implicit error:
```
diverging implicit expansion for type com.sksamuel.avro4s.SchemaFor[com.sksamuel.avro4s.schema.CoproductOfCoproductsField]
```

You can see it here in action

https://github.com/sksamuel/avro4s/compare/master...fedefernandez:ff/coproducts-in-fields?expand=1

Do you know why this could be happen? Thanks

Schm1tz1:
Hi @fedefernandez and @sksamuel,

I'm facing the sami issue in 2.0.2 on scala 2.11. I have a (nested) class that is used to store tree-like data structured which looks like the following:
```
import com.sksamuel.avro4s.{AvroSchema, SchemaFor}
import shapeless.{:+:, CNil}

case class DataWrapper(name: String,
                       dataType: String,
                       value: Option[String :+: Long :+: Double :+: CNil],
                       children: Option[List[DataWrapper]]
                      )


object AvroTest extends App {

  implicit val schemaFor = SchemaFor[DataWrapper]
  val newSchema = AvroSchema[DataWrapper]

  println(newSchema.toString(true))

}
```
This results in the same error:

> AvroTest.scala:13: error: diverging implicit expansion for type com.sksamuel.avro4s.SchemaFor[DataWrapper]
> starting with method optionSchemaFor in object SchemaFor
>   implicit val schemaFor = SchemaFor[DataWrapper]
>                                     ^
>AvroTest.scala:14: error: diverging implicit expansion for type com.sksamuel.avro4s.SchemaFor[DataWrapper]
> starting with method optionSchemaFor in object SchemaFor
>   val newSchema = AvroSchema[DataWrapper]

I have tried with the simple nested example from the README:
```
case class Recursive(payload: Int, next: Option[Recursive])

implicit val schemaFor = SchemaFor[Recursive]
val newSchema = AvroSchema[Recursive]
```

This leads to the same error. Any ideas/suggestions?

[...]

PR #311, opened 2019-05-15
Fixed various instance deriving problems relating to recursive types

2chilled:
This should fix at least #273. I guess the motivation for deciding for or against shapeless.Lazy was implicit resolution performance improvements. I did not benchmark it seriously, but I think it's still fast enough with shapeless.Lazy as default.

@sksamuel can you guess how long it takes to build a bugfix release from this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ credentials.sbt
 *.iml
 
 universe.avro
+
+# metals specific
+.metals
+.bloop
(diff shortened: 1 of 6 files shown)


sksamuel:
Can you try the 3.0.x snapshot from 
https://oss.sonatype.org/content/repositories/snapshots/com/sksamuel/avro4s
This should be fixed and is available for 2.11 and 2.22

PR #313, opened 2019-05-16
Fix #273

### LINKED ISSUE #273: Diverging implicit expansion for SchemaFor in Coproducts inside case classes

2chilled:
Fixes #273 by using shapeless.Lazy at the right places.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ credentials.sbt
 *.iml
 
 universe.avro
+
+# metals specific
+.metals
+.bloop
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/Github273.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/Github273.scala
@@ -1,6 +1,6 @@
 package com.sksamuel.avro4s.github
 
-import com.sksamuel.avro4s.{Decoder, Encoder, RecordFormat, SchemaFor}
+import com.sksamuel.avro4s.{Decoder, Encoder, SchemaFor}
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.Schema
 import org.scalatest.{FunSuite, Matchers}
@@ -14,12 +14,18 @@ class Github273 extends FunSuite with Matchers {
     implicitly[SchemaFor[Coproducts]].schema
     implicitly[SchemaFor[CoproductOfCoproductsField]].schema
     implicitly[SchemaFor[SealedTraitOfSealedTrait]].schema
+    implicitly[SchemaFor[DeeplyNested]].schema
+    implicitly[SchemaFor[MixedClassesAndObjects]].schema
     implicitly[Encoder[Coproducts]]
     implicitly[Encoder[CoproductOfCoproductsField]]
     implicitly[Encoder[SealedTraitOfSealedTrait]]
+    implicitly[Encoder[DeeplyNested]]
+    implicitly[Encoder[MixedClassesAndObjects]]
     implicitly[Decoder[Coproducts]]
     implicitly[Decoder[CoproductOfCoproductsField]]
     implicitly[Decoder[SealedTraitOfSealedTrait]]
+    implicitly[Decoder[MixedClassesAndObjects]]
+    implicitly[Decoder[DeeplyNested]]
   }
 
 }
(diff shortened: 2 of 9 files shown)


2chilled:
@sksamuel Would be nice to have a bugfix release including this patch asap as it's a show stopper for us upgrading to avro4s 2.x.x

2chilled:
Ok, this fails (again) for 2.11. Unfortunately I have zero motivation fixing it. Will release locally.

2chilled:
Addition: Support mixed object and class coproducts.

sksamuel:
This is great. I am also working on an upgrade at the moment.

sksamuel:
You might want to take a look at https://github.com/sksamuel/avro4s/tree/wip/magnolia which is what might become version 3.0. This should have your fixes applied.

sksamuel:
Do you want to try the 3.0 snapshot for 2.11 ?

2chilled:
Pretty busy this weekend, but sure I'll try 3.0 on Monday. Have you checked that something similar to Github273.scala tests is green for 3.0? Is there a plan for 3.0 release somewhere?

sksamuel:
Github273 has it's own unit test
https://github.com/sksamuel/avro4s/blob/master/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/Github273.scala

And is green locally and on travis.

sksamuel:
Master is now 3.0 btw.

sksamuel:
I say that, it's green locally, failed on travis. But it's only because _:+: is not allowed in the avro name. I'll fix that now.

sksamuel:
Green on travis now for 2.11 and 2.22
https://travis-ci.org/sksamuel/avro4s/builds/534123651

2chilled:
Sounds great! Let's forget about 2 and focus on 3 instead 👍

sksamuel:
I've added comments to most outstanding issues. Once I get some feedback that 3.0 is working I will deploy a release candidate. The only issue is recursive schemas, but they didn't work properly in 2.0 either.

2chilled:
Just tried 3.0.0-RC2. Unfortunately Magnolia seems to be in a loop, constantly emitting `magnolia: could not find any direct subtypes of trait CNil` while trying to derive an `Encoder` instance for our ADT. The process eventually terminates with `java.lang.OutOfMemoryError: GC overhead limit exceeded`. This week I don't have much time for providing a minimization of the problem but I'll see what I can do.

2chilled:
@sksamuel Are you sure you want typeclass methods to consume `NamingStrategy` from implicit scope? In my experience, the best compromise is to exclusively mark typeclass instances (no matter if typeclass is algebraically lawful or not) implicit.

sksamuel:
@2chilled I don't know what you mean by that. I need some way of passing `NamingStrategy` to the `ForSchema` macro.

sksamuel:
How did you get on ?

2chilled:
@sksamuel All I'm suggesting is to transform e.g. `def decode(value: Any, schema: Schema)(implicit naming: NamingStrategy = DefaultNamingStrategy): T` to `def decode(value: Any, schema: Schema, naming: NamingStrategy = DefaultNamingStrategy): T` Maybe even get rid of the default value, for the reasons noted [here](http://www.wartremover.org/doc/warts.html) 

For the moment I released the latest 2.x.x (2.12) with my patches on our local repo. I'm hoping to get mentioned 3.x.x issue reproduced by next weekend. Our production ADT is pretty deep with mixed `sealed trait`s and `shapeless.Coproduct`s, should hopefully not be too much work to minimize.

sksamuel:
I don't use wart remover myself so I assume this is the "wart" that you refer to:

```
ImplicitParameter
Implicit parameters as configuration often lead to confusing interfaces and can result in surprising inconsistencies.
```

It doesn't actually explain _why_ it is confusing. But that being said, I will see if I can remove the implicitness of the parameter. I had assumed I needed it to pass into the macro, but maybe I don't.

sksamuel:
I have removed the implicit naming strategy from SchemaFor, Encoder and Decoder. They remain in the "entry" points of AvroSchema, ToRecord and FromRecord. See what you think.

sksamuel:
Closing this as I believe it to be fixed in the 3.0 series and haven't heard otherwise.

PR #315, opened 2019-05-17
Use optional decoder to read non optional fields

ivanobulo:
I want to be able read fields that declared as non-optional in Avro schema utilizing `optionDecoder`.
This is common case in my project when I migrate older model to the new one where certain fields become optional.
With this change I could read "legacy" records into the new schema where certain fields are now optional.
The problem with existing implementation is that in such case it throws AvroRuntimeException with "Not a union: " message.


sksamuel:
So you want to have optional fields that are read in from a schema where the field is not marked as optional.

ivanobulo:
exactly!

sksamuel:
I've merged this and will apply it manually to the 3.0 pre release branch

sksamuel:
Can you try the 3.0.x snapshot from 
https://oss.sonatype.org/content/repositories/snapshots/com/sksamuel/avro4s/avro4s-core_2.12/
This should be fixed and is available for 2.11 and 2.22

PR #317, opened 2019-05-21
Add default values for seqs and maps

alexjg:
I noticed that for some reason the test for default arrays and maps had been omitted. I was running into issues generating schemas with these kind of default values so I've had  a stab at implementing the feature here.

For some reason this broke the `CoproductEncoderTest`. I couldn't figure out exactly what the problem was but making the type name  check slightly more flexible in `SchemaHelper` fixed the test.


alexjg:
I'm a little confused by the test failures here. I can't see how they're related to the work in the PR

sksamuel:
The failure is the snapshot deployer. I'll fix that as it's unrelated to your work.

PR #319, opened 2019-05-21
Optional unions with default values set to None

alexjg:
I ran into a problem where setting the default value of an `Option` field caused an error. This was because of the way the current behaviour which allows a default value of `null` to generate a union schema was implemented.

I've fixed the behaviour so that setting a default value of `None` renders a `null` in the schema. However, this might not be a complete implementation because I can see an argument that if the default value is set to `None` then reading an older record (i.e one written before the field existed and hence requiring the default value) should return `None`, and I _think_ that at the moment it would read `null`.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
@@ -218,7 +218,7 @@ object SchemaFor {
     }
 
     // if our default value is null, then we should change the type to be nullable even if we didn't use option
-    val schemaWithPossibleNull = if (default == Some(null)) {
+    val schemaWithPossibleNull = if (default == Some(null) && schema.getType != Schema.Type.UNION) {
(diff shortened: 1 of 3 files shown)


avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
```diff
@@ -217,7 +218,7 @@ object SchemaFor {
     }
 
     // if our default value is null, then we should change the type to be nullable even if we didn't use option
-    val schemaWithPossibleNull = if (encodedDefault == JsonProperties.NULL_VALUE) {
+    val schemaWithPossibleNull = if (default == Some(null)) {
```
sksamuel: Maybe this should just be `if (default == Some(null) && schema.getType != Schema.Type.Union`
Because the use case we are trying to catch is that the user has done `case class Foo(a: String = null)`. If they used None as the default, then they must have an Option on the left hand side of that, so the union will be generated anyway.

alexjg: Good shout, updated.

sksamuel:
You can try the snapshot to see if it's working for you.

PR #320, opened 2019-05-22
Optional unions

alexjg:
This is another part of defaults for `map` and `array` types that I missed. Specifically, in cases where the default is a map or array it wasn't moving the default value to the head of the type list in the union.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaHelper.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaHelper.scala
@@ -95,6 +95,8 @@ object SchemaHelper {
       case _: Double => Schema.Type.DOUBLE
       case _: Array[Byte] => Schema.Type.BYTES
       case _: GenericData.EnumSymbol => Schema.Type.ENUM
+      case _: java.util.ArrayList[_] => Schema.Type.ARRAY
+      case _: java.util.LinkedHashMap[_,_] => Schema.Type.MAP
       case JsonProperties.NULL_VALUE => Schema.Type.NULL
       case other => other
     }
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaHelper.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaHelper.scala
@@ -95,8 +95,8 @@ object SchemaHelper {
       case _: Double => Schema.Type.DOUBLE
       case _: Array[Byte] => Schema.Type.BYTES
       case _: GenericData.EnumSymbol => Schema.Type.ENUM
-      case _: java.util.ArrayList[_] => Schema.Type.ARRAY
-      case _: java.util.LinkedHashMap[_,_] => Schema.Type.MAP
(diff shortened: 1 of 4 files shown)


avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaHelper.scala
```diff
@@ -95,6 +95,8 @@ object SchemaHelper {
       case _: Double => Schema.Type.DOUBLE
       case _: Array[Byte] => Schema.Type.BYTES
       case _: GenericData.EnumSymbol => Schema.Type.ENUM
+      case _: java.util.ArrayList[_] => Schema.Type.ARRAY
+      case _: java.util.LinkedHashMap[_,_] => Schema.Type.MAP
```
sksamuel: This should probably just be Map right?

avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaHelper.scala
```diff
@@ -95,6 +95,8 @@ object SchemaHelper {
       case _: Double => Schema.Type.DOUBLE
       case _: Array[Byte] => Schema.Type.BYTES
       case _: GenericData.EnumSymbol => Schema.Type.ENUM
+      case _: java.util.ArrayList[_] => Schema.Type.ARRAY
```
sksamuel: Could be any collection type ?

avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaHelper.scala
alexjg: No because by this point it's been through the `DefaultResolver`, which turns `Map` into a `LinkedHashMap` because, as far as I can tell, that's what Avro expects for default maps. Likewise for the `ArrayList`.

sksamuel: Yes fair enough, although it does mean changing one could leave to bugs down the line in the other.

sksamuel: Also, is it in the spec that avro expects a linked map ?

alexjg: No there's no spec. I was just getting errors using a `Map` as a default value and so I loaded up a schema from JSON with a default value and had a look at what the type was. However, now that I think about it the error may have been unrelated to the collection type. I'll have a crack at using `Map` and `Seq` here and in the `DefaultResolver` and see what happens.

sksamuel: It would just seem odd for anyone to rely on linked hash map, although with avro I wouldn't be surprised.

alexjg: Yup, look like I was mistaken. I've updated it so that defaults can be any kind of collection.

sksamuel:
Once it goes green we can merge.

PR #324, opened 2019-05-24
Major refactoring for NamingStrategy

ivanobulo:
It is now also possible to apply NamingStrategy implicitly:

```scala
    implicit val naming: SnakeCase.type = SnakeCase
    val schema = SchemaFor[NamingTest]
    val decoder = Decoder[NamingTest]
```

I've added test case for that.


ivanobulo:
Encoder naming test is still failing because I haven't implemented for one yet. But naming tests for schema and decoder pass.

ivanobulo:
@sksamuel this is done. all unit tests pass. idk why the CI build is failed.

ivanobulo:
After thinking more about this change I feel like the only benefit of this change is backward compatibility but at increased complexity of the code.
If you feel the same go ahead and just close this PR.

sksamuel:
The idea of keeping the naming strategy with the schema is quite powerful
though.

ivanobulo:
I'll try to look at this with the fresh angle. I got much better understanding of the code after this PR so I hope I can come up with smth less complex to address this problem.

sksamuel:
So where did we get up to with this.
I think deciding on the naming strategy is the only thing stopping me releasing 3.0 final.

ivanobulo:
I didn't have time to reflect on this. Let's continue with your original solution.

PR #332, opened 2019-06-11
Fix incorrect name lookup with enums

alexjg:
Ran into an issue where enums with the `@AvroName` annotation were failing to decode. For some reason it seems to only be replicated by a round trip through the encoder and decoder. I've attached a test which reproduces the issue but it may be in the wrong place. (given that it's in the `decoder` test folder).


sksamuel:
:+1:

PR #335, opened 2019-06-21
Adds refined module

### LINKED ISSUE #327: Add refined module
I think it would be cool if there was a [refined](https://github.com/fthomas/refined) module, providing avro4s typeclass instances for refined types. 

I have some experience with this so would be happy to give it a go.

I envision it being setup same way as the cats module.

Let me know what you think!

lacarvalho91:
Closes #327 

Let me know if I should comment out the `avro4s-refined` sbt module like the cats one is, or if you'd like the cats one uncommented (if the only thing missing for the cats one is tests I'd be happy to add those to this PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -732,11 +732,21 @@ This will result in a schema where both `BigDecimal` quantities have their own s
 
 ## Cats Support
 
-If you use cats in your domain objects, then Avro4s provides a cats module with schemas, encoders and decoders for some cats typs.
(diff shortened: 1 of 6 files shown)


lacarvalho91:
Not sure whats going on with the Travis build 😭

sksamuel:
Yeah ignore that I need to fix that.

sksamuel:
:+1:

PR #341, opened 2019-06-25
fix #339 adds support for immutable Seq and more

### LINKED ISSUE #339: no support for immutable Seq?
Scala `2.12`
Avro4s `3.0.0-RC1`

I was just playing around with the Pizza examples and adding `import scala.collection.immutable.Seq` seems to cause schema generation to fail

```
could not find implicit value for parameter schemaFor: com.sksamuel.avro4s.SchemaFor[test.Pizza]
[error]     val schema = AvroSchema[Pizza]
```

francisdb:
Probably first needs a review

* For schema generation any `Iterable` is now accepted and yields `Array` in the schema
* Encoders/Decoders need some more explicit handling becuase the low priority implicit is a bit less performant

I have also seen that Scala 2.13 fails because of missing Magnolia dependency.
Might need some more updates for 2.13 because of the collections changes
https://github.com/scala/scala-collection-compat

Got some inspiration here:
https://github.com/playframework/play-json/blob/dd7c7d1d9c9f329829809bd13e0dccd058f30c1c/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala#L395
https://github.com/playframework/play-json/blob/dd7c7d1d9c9f329829809bd13e0dccd058f30c1c/play-json/shared/src/main/scala/play/api/libs/json/Reads.scala#L229

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/Decoder.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/Decoder.scala
@@ -212,6 +212,8 @@ object Decoder {
 
   implicit def listDecoder[T](implicit decoder: Decoder[T]): Decoder[List[T]] = seqDecoder[T](decoder).map(_.toList)
 
+  implicit def mutableSeqDecoder[T](implicit decoder: Decoder[T]): Decoder[scala.collection.mutable.Seq[T]] = seqDecoder[T](decoder).map(_.to[scala.collection.mutable.Seq])
+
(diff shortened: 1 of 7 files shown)


francisdb:
How do I fix this travis `key not found: OSSRH_USERNAME`?

sksamuel:
I need to fix the snapshot deploy.sh

francisdb:
Now we have to wait for magnolia for scala 2.13 to become available so we can validate this on 2.13

sksamuel:
:+1:

PR #351, opened 2019-07-17
Issue 350 - encoding a record with a union annotated with @AvroNamespace

bobbyrauchenberg:
This is a fix for an issue where an exception got raised when encoding records that contained unions, annotated with `@AvroNamespace`

The code that combined the schemas for the types described by the union was using a `foreach`, which didn't affect the schema that was subsequently being encoded

I have replaced that `foreach` with a `foldLeft`


sksamuel:
:+1:

Issue #352, opened 2019-07-19
User defined classes can't be used as defaults in a Union


bobbyrauchenberg:
Currently it's not possible to use a user defined class as a default value in a Union

For this ADT
```
sealed trait Dog
case object HotDog extends Dog
case class CupcatDog(name: String) extends Dog
```

For this case class
```
case class Cupcat(dog: Dog = HotDog)
```

This test
```
"support default value of an arbitrary type" in {
      val schema = AvroSchema[Cupcat]
      val expected = new org.apache.avro.Schema.Parser().parse(getClass.getResourceAsStream("/optional_default_values.json"))
      schema.toString(true) shouldBe expected.toString(true)
    }
```
Throws 
```
Unknown datum class: class com.sksamuel.avro4s.schema.CupcatDog
org.apache.avro.AvroRuntimeException: Unknown datum class: class com.sksamuel.avro4s.schema.CupcatDog
	at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:87)
	at org.apache.avro.util.internal.JacksonUtils.toJsonNode(JacksonUtils.java:48)
	at org.apache.avro.Schema$Field.<init>(Schema.java:530)
	at com.sksamuel.avro4s.SchemaFor$.com$sksamuel$avro4s$SchemaFor$$buildField(SchemaFor.scala:246)
	at com.sksamuel.avro4s.SchemaFor$$anon$14.$anonfun$schema$9(SchemaFor.scala:305)
	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:244)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:39)
	at scala.collection.TraversableLike.flatMap(TraversableLike.scala:244)
	at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:241)
	at scala.collection.AbstractTraversable.flatMap(Traversable.scala:108)
	at com.sksamuel.avro4s.SchemaFor$$anon$14.schema(SchemaFor.scala:282)
	at com.sksamuel.avro4s.AvroSchema$.apply(AvroSchema.scala:21)
[...]

PR #354, opened 2019-07-20
Issue #352 - User defined classes can't be used as defaults in a Union

bobbyrauchenberg:
- allows union defaults to be user defined case classes / case objects
- handles correctly reordering the schema as per the spec
- case class defaults are json objects
- case object defaults are empty json object


avro4s-core/src/test/scala/com/sksamuel/avro4s/schema/DefaultValueSchemaTest.scala
```diff
@@ -53,29 +53,40 @@ class DefaultValueSchemaTest extends WordSpec with Matchers {
...
-                         isFemale: Option[Boolean] = Some(false),
-                         length: Option[Double] = Some(6.2),
-                         timestamp: Option[Long] = Some(1468920998000l),
-                         address: Option[Map[String, String]] = Some(Map(
-                           "home" -> "sammy's home address",
-                           "work" -> "sammy's work address"
-                         )),
-                         traits: Option[Seq[String]] = Some(Seq("Adventurous", "Helpful")),
-                         favoriteWine: Option[Wine] = Some(Wine.CabSav))
+                                 age: Option[Int] = Some(21),
```
bobbyrauchenberg: sorry, this stuff got formatted by my IDE

sksamuel:
Sorry for the delaying in reviewing.

PR #359, opened 2019-08-01
Correctly decode tuples that contain a Seq

env0der:
Otherwise decoding fails with `org.apache.avro.AvroRuntimeException: Not an array`. I guess this would also fix decoding of tuples that contain other complex types.


env0der:
Note: the CI build has failed because it could not install oraclejdk8. I'm not sure what to do with that issue.

sksamuel:
Thanks for the fix !

Issue #364, opened 2019-08-18
Default Enum values don't work


bobbyrauchenberg:
Currently if you try to provide a default value for either a Scala Enumeration, or a sealed trait enum, schema creation doesn't work

e.g. for 
```
object Colours extends Enumeration {
  val Red, Amber, Green = Value
}
case class ScalaEnumsWithDefault(colours: Colours.Value = Colours.Red)
```
then
```AvroSchema[ScalaEnumsWithDefault]```
gives
```Unknown datum class: class scala.Enumeration$Val
org.apache.avro.AvroRuntimeException: Unknown datum class: class scala.Enumeration$Val```

PR #365, opened 2019-08-18
Fix to allow defaults for enum values

bobbyrauchenberg:
Fix for Issue #364 
* Handles scala enums
* Handles sealed trait enums

Also handles defaults for Option


bobbyrauchenberg:
this (pretty horrible) line ```case e: Enumeration$Val => CustomDefaultResolver.customScalaEnumDefault(value)```

i've added to `DefaultResolver`, doesn't seem to compile against scala 2.13

sksamuel:
Great work. Thanks again.

PR #374, opened 2019-09-18
Small PR to add a test around sort priority and defaults

bobbyrauchenberg:
Tests that the @AvroSortPriority annotation, does not break the reordering of the schema fields that happens when we add a default union value

Also fixes a small typo in test class naming


sksamuel:
:+1:

PR #376, opened 2019-09-19
fixes broken test for issue 202

bobbyrauchenberg:
This is a fix for the test in github-202

There is no issue with Avro4s, just some weirdness with how aliases work in Avro

If you add an alias to a single field, you need to alias all fields and alias on the record level too

I've added a pure java api version of the test too, as i think it makes the solution very explicit :)


sksamuel:
👍

PR #377, opened 2019-09-26
bump Scala versions, modernize Travis-CI config

SethTisue:
I think it's a bit more standard (and more parallel) to put the matrix of Scala versions in `.travis.yml`, rather than using `+test` and `+publish`. it's nice to see at a glance in the Travis-CI results which versions are failing

btw it seems a bit suspicious to me that you have two JDKs in the Travis-CI matrix but you run `snapshot_deploy.sh` from both, aren't you double-publishing snapshots? I didn't check or change this

was it intentional that you weren't building/testing on JDK 8, only 10 and 11? I changed it to 8 and 11, but perhaps that's wrong

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,14 @@
-sudo: false
-
 language: scala
 
 scala:
-- 2.12.8
+- 2.12.10
+- 2.13.1
 
 script:
-- sbt +test
+- sbt ++$TRAVIS_SCALA_VERSION test
 
 jdk:
-- openjdk10
-- oraclejdk11
+- openjdk8
+- openjdk11
 
 after_success: "./snapshot_deploy.sh"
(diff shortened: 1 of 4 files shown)


SethTisue:
why am I submitting this? it's because I'm seeing avro4s fail in the Scala 2.13 community build. I think you may be running afoul of some 2.13.0 vs 2.13.1 difference. I want to see if it fails in this PR too

SethTisue:
gah, the 2.13.1/openjdk8 job failed because of a transient network problem, but I don't have the rights to hit "Restart" on it

SethTisue:
yup, same error here.

but then I revised it to add:

```
-case class AvroEnumDefault(default: Any)
+case class AvroEnumDefault(default: Any) extends StaticAnnotation
```

and the problem went away. that seems like what was intended anyway, so I guess this was a progression in 2.13.1, that the compiler now notices

anyway, feel to merge this whole PR, or just close it and cherry-pick the parts you like instead

sksamuel:
great stuff thanks

PR #384, opened 2019-10-11
FromRecord should decode using reader schema

r-glyde:
I think `FromRecord` should be decoding using the reader schema (for the class being decoded to) rather than the schema of the record being decoded. Currently with more than basic classes, `T` needs the same fields as the writer.

Sorry, if this is wrong but I think it makes sense unless I've misunderstood what `FromRecord` is supposed to be used for?

The test seemed to fit in with the decoding tests but let me know if there is a better place for it.


sksamuel:
Good spot thanks.

PR #388, opened 2019-10-23
ISSUE-387 | Fixed LocalTime Encoder to properly convert to milliseconds

rbnks:
~~Fixed the Encoder for LocalTime objects to properly convert from nanoseconds to millisseconds. Nano second values are truncated to the millisecond precision. LocalTime.now() produces millisecond precision values by default so the schema and encoder/decoder are inline with default functionality of the LocalTime object even though LocalTime does allow for a higher level of precision.~~

In Java8 the default functionality of LocalTime.now() produces millisecond precision, however, in Java11 it produces microsecond precision.  I change the logical type of LocalTime to be `time-micros` and updated the encoder and tests accordingly.

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/schema/DateSchemaTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/schema/DateSchemaTest.scala
@@ -1,7 +1,7 @@
 package com.sksamuel.avro4s.schema
 
 import java.sql.{Date, Timestamp}
-import java.time.{Instant, LocalDate, LocalTime}
+import java.time.{Instant, LocalDate, LocalDateTime, LocalTime}
 
 import com.sksamuel.avro4s.AvroSchema
 import org.scalatest.{FunSuite, Matchers}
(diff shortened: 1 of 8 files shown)


avro4s-core/src/test/scala/com/sksamuel/avro4s/schema/DateSchemaTest.scala
```diff
@@ -30,9 +30,9 @@ class DateSchemaTest extends FunSuite with Matchers {
   }
 
   test("generate time logical type for LocalDateTime") {
-    case class LocalTimeTest(time: LocalTime)
+    case class LocalDateTimeTest(time: LocalDateTime)
     val expected = new org.apache.avro.Schema.Parser().parse(getClass.getResourceAsStream("/localdatetime.json"))
-    val schema = AvroSchema[LocalTimeTest]
+    val schema = AvroSchema[LocalDateTimeTest]
     schema.toString(true) shouldBe expected.toString(true)
```
rbnks: When I fixed the schema test for LocalTime, I noticed the test for LocalDateTime was the test for LocalTime copied and pasted.  I fixed this test as well.

sksamuel:
Do you think the LocalTimeDecoder should be updated to handle both situations? So that anyone upgrading doesn't get an error on data already written.

rbnks:
The LocaTimeDecoder already was set up to handle both cases.  It will successfully parse micro and milliseconds based on the logical type in the schema.  https://github.com/sksamuel/avro4s/blob/863b8a628d03acc430ce4020dbce7a3d3da8507a/avro4s-core/src/main/scala/com/sksamuel/avro4s/Decoder.scala#L146

sksamuel:
Excellent fix thanks.

PR #390, opened 2019-10-24
ISSUE-389 | Add support for OffsetDateTime

rbnks:
Added support for OffsetDateTime objects. Created a custom logical type 'datetime-with-offset' that is backed by string. OffsetDateTimes are encoded by converting them to a String using the DateTimeFormatter.ISO_OFFSET_DATE_TIME . Example Avro schema can be found in test/resources/github/github_389.json.


sksamuel:
Really cool.

Issue #385, opened 2019-10-14
Maybe 'Record' should not extend both 'GenericRecord' and 'SpecificRecord'?


yeryomenkom:
Hi!
We (hello @maxstreese) use avro4s together with [confluent_serdes](https://github.com/confluentinc/schema-registry/tree/master/avro-serde/src/main/java/io/confluent/kafka/streams/serdes/avro) in our project.
And we found weird serializing performance issue today.
That's how we create serdes (simplified a bit) :
```
private def createSerde[T >: Null](genericSerde: GenericAvroSerde, format: RecordFormat[T]): Serde[T] = {
    val serde: Serde[T] = Serdes.fromFn(
      serializer = (topic, value) => {
        if (value == null) null
        else genericSerde.serializer().serialize(topic, format.to(value))
      },
      deserializer = ???
    )

    serde
  }
```
And, if we will try to use this serde, for 50K records we will get the next timings:
```
Serde serialization. Time: 3.671 seconds.
Serde deserialization. Time: 0.409 seconds.
```
The reason lives inside the implementation of `AbstractKafkaSerializer` [link](https://github.com/confluentinc/schema-registry/blob/master/avro-serializer/src/main/java/io/confluent/kafka/serializers/AbstractKafkaAvroSerializer.java#L88).
As you can see, it always treats container value as `SpecificRecord`, since `Record` is indeed a proper implementation of `SpecificRecord`.
If we will use workaround and just wrap `Record` in the `WrappedRecord`:
```
class WrappedRecord(val record: Record) extends GenericRecord {
  override def put(key: String, v: Any): Unit = record.put(key, v)
  override def get(key: String): AnyRef = record.get(key)
  override def put(i: Int, v: Any): Unit = record.put(i, v)
  override def get(i: Int): AnyRef = record.get(i)
  override def getSchema: Schema = record.getSchema
}
```
the timings will be the the following:
```
Serde serialization. Time: 0.683 seconds.
Serde deserialization. Time: 0.409 seconds.
```

What do you think about all this situation? Do we do something wrong? Or maybe we should use another workaround?

Thanks!

sksamuel:
The implement of Record is `case class ImmutableRecord(schema: Schema, values: Vector[AnyRef]) extends Record`
So I guess the vector is slow with indexed lookup and if it's using the specific record (get by index) then it might be slow. In your wrapper it might be using the get by key? 
We could try changing the record implementation to using an array.

yeryomenkom:
[...]

PR #392, opened 2019-10-27
Improve performance of records serialization

yeryomenkom:
It is rather a continuation of the [discussion](https://github.com/sksamuel/avro4s/issues/385) then a proper PR. 
As you can see there is still a significant difference between specific and generic serialization even after using `Array` instead of `Vector`. 
Honestly speaking I don't have clear understanding of `SpecificDatumWriter` and `GenericDatumWriter`. Maybe something is wrong there...

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/encoder/StructEncoderTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/encoder/StructEncoderTest.scala
@@ -1,15 +1,65 @@
 package com.sksamuel.avro4s.record.encoder
 
-import com.sksamuel.avro4s.{AvroSchema, DefaultFieldMapper, Encoder, ImmutableRecord}
+import java.io.ByteArrayOutputStream
+
+import com.sksamuel.avro4s.TimeMeasurementUtils._
+import com.sksamuel.avro4s.{AvroSchema, DefaultFieldMapper, Encoder, ImmutableRecord, Record}
+import org.apache.avro.Schema
+import org.apache.avro.generic.{GenericDatumWriter, GenericRecord}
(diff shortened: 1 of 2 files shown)


avro4s-core/src/test/scala/com/sksamuel/avro4s/record/encoder/StructEncoderTest.scala
```diff
@@ -22,5 +72,34 @@ class StructEncoderTest extends WordSpec with Matchers {
...
+        (0 until n) foreach { _ => emulateEncodingForSpecificRecord(immutableRecordArray) }
+      }
+
+      logTime("ImmutableRecord. Generic serialization") {
+        (0 until n) foreach { _ => emulateEncodingForGenericRecord(immutableRecord) }
+      }
+      logTime("ImmutableRecord. Specific serialization") {
+        (0 until n) foreach { _ => emulateEncodingForSpecificRecord(immutableRecord) }
+      }
+    }
```
yeryomenkom: Here is my results:
```
ImmutableRecordArray. Generic serialization Time: 0.447 seconds.
ImmutableRecordArray. Specific serialization Time: 1.615 seconds.
ImmutableRecord. Generic serialization Time: 0.526 seconds.
ImmutableRecord. Specific serialization Time: 1.804 seconds.
```

sksamuel:
I wonder if this is related https://avro.apache.org/docs/current/gettingstartedjava.html#Beta+feature:+Generating+faster+code

yeryomenkom:
> I wonder if this is related https://avro.apache.org/docs/current/gettingstartedjava.html#Beta+feature:+Generating+faster+code

Sorry for late reply. I had no enough time to work on it. I will investigate this possibility and maybe dive a bit deeper into serialization process.

PR #395, opened 2019-11-06
Issue 394

rbnks:
Changed object to be implicit to resolve compiler errors in projects that depend on this project.


sksamuel:
Out of interest, why does the logical type need to be implicit? It looks like it's just used explicitly ?

Issue #406, opened 2019-12-03
Non-derivation api


xplosunn:
I can't seem to find it in the code (maybe I'm overlooking something) but there doesn't seem to be a way to use avro4s implicits without the auto-derivation. Like when using circe, I avoid any sort of derivation because I want the compile-time checking that every definition is implemented, but I don't like that renaming of a field on a case class is a breaking change. 

Is this implemented? Is it something that would make sense for this library?

I'd be willing to contribute.

sksamuel:
Can you give me an example? Do you mean by just creating the implicits manually ?

`implicit val xdecoder = Decoder[X] { ... }`

And yes, please feel free to contribute!

xplosunn:
Yes. It would mean to allow to create the implicits manually. But then, for schema, encoders and decoders to be consistent I think they would need to be created in a single instance.

It would probably lead to the following trait:

```
trait SchemaCodec[T] { self =>
  def schemaFor: SchemaFor[T]
  def encoder: Encoder[T]
  def decoder: Decoder[T]
}
```

And then in the companion object there would contain instances for primitives and collections and helper methods to create records. 

If the concept looks good to you, I'll proceed with an implementation, @sksamuel

sksamuel:
That sounds good, but how would you pass those instances without changing
the existing api ?

xplosunn:
You can work with implicit `SchemaCodec` instances only, the existing constructs are passed explicitly when building a `SchemaCodec`.

Maybe it's more obvious once I open up a PR.

sksamuel:
Thanks for the PR. So what I was curious was what would you use the `SchemaCodec` for once you've pulled together an instance for something?

xplosunn:
As mentioned in the pull request, I'm no longer using avro at the moment so I'm not sure when / if I'll come back to this effort.

PR #409, opened 2019-12-08
Implement SchemaCodec

xplosunn:
It's still missing tests and it's only done for core, but hopefully it's enough to start discussion.

#406


xplosunn:
I am no longer using avro at the moment so I'm not sure when / if I'll come back to implement this. Sorry.

PR #410, opened 2019-12-11
recursively resolve default values

larochef:
if a default value is a case class with default values, they were not
correctly handled.

Use the default resolver back to resolve the default value in the
default value recursively

Known issue: at some point, we need the type name if the schema is a
union. The name used here does't resolve the correct name if an
@AvroName annotation is used on the class. Same applies on the field
name, no FieldMapper is applied.

Also, in `CustomUnionDefault` some code left me a bit puzzled, so I changed it to what seemed the right thing

```scala
        CustomUnionDefault(trimmedClassName(p), parse(write(p)).extract[Map[String, Any]].mapValues {
          case (name, b: BigInt) if b.isValidInt => b.intValue: Any
          case (name, b: BigInt) if b.isValidLong => b.longValue: Any
        }
```
changed to:

```scala
        CustomUnionDefault(trimmedClassName(p), parse(write(p)).extract[Map[String, Any]].map {
          case (name, b: BigInt) if b.isValidInt => name -> b.intValue
          case (name, b: BigInt) if b.isValidLong => name -> b.longValue
        }
```
The unit test only checks that no exception is thrown when generating the schema

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/CustomDefaults.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/CustomDefaults.scala
@@ -5,6 +5,7 @@ import magnolia.{SealedTrait, Subtype}
 import org.json4s.native.JsonMethods.parse
 import org.json4s.native.Serialization.write
 import org.apache.avro.Schema
+import org.apache.avro.Schema.Type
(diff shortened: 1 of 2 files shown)


sksamuel:
awesome thanks

joan38:
Thanks

PR #412, opened 2019-12-12
Use SAM type to instantiate default SchemaFor

joan38:
This also solves the issue that `SchemaFor` is not serializable since `Schema` is instantiated eagerly and therefore needs to be serializable.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaFor.scala
@@ -51,26 +51,19 @@ object SchemaFor {
 
   def apply[T](implicit schemaFor: SchemaFor[T]): SchemaFor[T] = schemaFor
 
-  /**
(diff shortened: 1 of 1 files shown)


sksamuel:
Nice

chenharryhua:
Might I ask why delete `const` method?

joan38:
For 2 reasons:
- With SAM types you achieve the same without `const` (that was the reason SAM has been invented)
- `const` makes the `SchemaFor` unSerializable because `Schema` is instantiated eagerly and passed to `const`. `Schema` is not `Serializable`.

chenharryhua:
Thanks, @joan38 .

svroonland:
Just a friendly pointer that this change was not backwards compatible (for me at least) but was released as a patch release. Is `avro4s` following semantic versioning?

joan38:
That's indeed not binary compatible.
What issue did you experience?

svroonland:
@joan38 I was using `SchemaFor.const`.

joan38:
I see it was not private and we did no deprecation cycle.

Issue #413, opened 2019-12-14
Bug in the new schema enum default code


stevewampler:
The new schema enum default code is putting the `default` field in the wrong location in the schema.
For example, the `EnumSchemaTest."support default scala enum"` test case generates the following schema where the `default` field is placed outside of the `enum` JSON object:
```
{
  "type": "record",
  "name": "ScalaEnumsWithDefault",
  "namespace": "com.sksamuel.avro4s.schema",
  "fields": [
    {
      "name": "colours",
      "type": {
        "type": "enum",
        "name": "Colours",
        "namespace": "com.sksamuel.avro4s.schema",
        "symbols": [
          "Red",
          "Amber",
          "Green"
        ]
      },
      "default": "Red"
    }
  ]
}
```
According to https://avro.apache.org/docs/current/spec.html#Enums, the `default` field should be at the same level as the `symbols` field:
```
{
  "type": "record",
  "name": "ScalaEnumsWithDefault",
  "namespace": "com.sksamuel.avro4s.schema",
  "fields": [
    {
      "name": "colours",
      "type": {
        "type": "enum",
        "name": "Colours",
        "namespace": "com.sksamuel.avro4s.schema",
        "symbols": [
          "Red",
          "Amber",
          "Green"
        ],
        "default": "Red"
      }
    }
  ]
}
```
Below are two tests that check the compatibility of the above schema when adding a new enum value `Orange`. The first test puts the `default` field in the same location as avro4s 3.0.4 and fails the compatibility test. The second test puts the `default` field next to the `symbols` field and succeeds:
```import com.sksamuel.avro4s.AvroSchema
import org.apache.avro.SchemaCompatibility
import org.apache.avro.SchemaCompatibility.SchemaCompatibilityType
import org.junit.jupiter.api.Assertions._
import org.junit.jupiter.api.Test

class Avro4sEnumDefaultTest {
  object Colours extends Enumeration {
    val Red, Amber, Green = Value
  }

  case class ScalaEnumsWithDefault(colours: Colours.Value = Colours.Red)

  // This test uses the AVRO SchemaCompatibility.checkReaderWriterCompatibility method to check the compatibility of two
  // versions of an enum schema with a default enum field in the location produced by avro4s 3.0.4.
  // In the first version of the schema, the
  // enum values are "Red", "Amber", and "Green". In the second version, "Orange" is added to the enum values.
  //
  // The test shows that the two schemas versions are not actually compatible because the "default" enum value field
[...]

PR #420, opened 2020-01-03
Improved support for Enum Defaults

stevewampler:
This PR enables an avro4s user to specify a default enum value for a scala or java enum. Enum defaults enable AVRO enums to be backwards compatible (see https://avro.apache.org/docs/current/spec.html#Enums).

Note that an enum default is not the same as a field default. AVRO enum defaults are specified as part of the enum's type definition:
```json
{
   "type" : "enum",
   "name" : "Colours",
   "symbols" : [ "Red", "Amber", "Green" ],
   "default": "Amber"
}
```
whereas field defaults are defined at the field level and may differ from the enum default:
```json
{
  "type" : "record",
  "name" : "Car",
  "fields" : [ {
    "name" : "colour",
    "type" : {
      "type" : "enum",
      "name" : "Colours",
      "symbols" : [ "Red", "Amber", "Green" ],
      "default": "Amber"
    },
    "default": "Red"
  } ]
}
```
To enable users to specify an enum's default, this PR adds two new methods--one for Java enums and one for Scala enums--that add the default enum value/symbol to the enum's schema:
```scala
object JavaEnumSchemaFor extends EnumSchemaFor {
  def apply[E <: Enum[_]](default: E)(implicit tag: ClassTag[E]): SchemaFor[E] = SchemaFor.javaEnumSchemaFor.map[E](addDefault(default))
}

object ScalaEnumSchemaFor extends EnumSchemaFor {
  def apply[E <: scala.Enumeration#Value](default: E)(implicit tag: TypeTag[E]): SchemaFor[E] = SchemaFor.scalaEnumSchemaFor.map[E](addDefault(default))
}
```
The `addDefault` method is defined in the `EnumSchemaFor` trait:
```scala
trait EnumSchemaFor {
  protected def addDefault[E](default: E)(schema: Schema): Schema = SchemaBuilder.
    enumeration(schema.getName).
    namespace(schema.getNamespace).
    defaultSymbol(default.toString).
    symbols(schema.getEnumSymbols.asScala.toList:_*)
}
```
It copies the enum's schema and adds to it the default enum symbol.

For example, to generate an AVRO enum schema--with a default enum value--for the following scala enum:
```scala
object Colours extends Enumeration {
   val Red, Amber, Green = Value
}
 ```
you use `ScalaEnumSchemaFor` as follows:
```scala
implicit val schemaForColours: SchemaFor[Colours.Value] = ScalaEnumSchemaFor[Colours.Value](default = Colours.Amber)
```
The resulting AVRO schema will be:
```json
{
   "type" : "enum",
   "name" : "Colours",
   "symbols" : [ "Red", "Amber", "Green" ],
   "default": "Amber"
}
```

Changes:

* Added `JavaEnumSchemaFor` and `ScalaEnumSchemaFor` to allow users to specify a default enum value when creating an `enum` schema from Java or Scala enums.
* Added all of the various permutations of enums with default values and default fields to `EnumSchemaTest`. I inlined the expected schema json strings to make it easier to read the tests and the expected results.
* Created an `EnumSchemaCompatibilityTest` to demonstrate the enum backwards-compatibility issue.
* Added info to the README about enums and enum defaults.

#413


sksamuel:
Great work. Is it worth supporting AvroEnumDefault on enums as well in case you don't want to make the implicit ?

stevewampler:
@sksamuel That's a great idea. Let me see if I can make that work ...

stevewampler:
@sksamuel I added code to `SchemaFor.scalaEnumSchemaFor` to handle `@AvroEnumDefault` annotations for scala enums. While I was at it, I added changed the code to handle `@AvroName`, `@AvroNamespace`, and `@AvroProp` for scala enums as well. The existing code for those annotations wasn't working.

I also changed `SchemaFor.javaEnumSchemaFor`. The scala `@AvroEnumDefault` didn't work for java enums (unless I'm missing something), so I created a java annotation called `@AvroJavaEnumDefault`. I also created `@AvroJavaName`, `@AvroJavaNamespace`, and `@AvroJavaProp` that can be applied to java enums as well.

sksamuel:
This is amazing. If I can be a tiny bit nit picky, could we re-use the normal AvroName for AvroJavaName and so forth (or is there a reason it can't be done ?)

stevewampler:
Thanks @sksamuel. I was hoping to use the `@AvroName`, but it's targeted for `ElementType.FIELD` only:
```java
@Target(ElementType.FIELD)
@Retention(RetentionPolicy.RUNTIME)
public @interface AvroName {
  String value();
}
```
and not `ElementType.TYPE`. Some of the other AVRO annotations, such as `@AvroDoc` are targeted for both fields and types:
```java
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.TYPE, ElementType.FIELD })
public @interface AvroDoc {
  String value();
}
```
Without the `ElementType.TYPE`, I receive `Error:(9, 1) java: annotation type not applicable to this kind of declaration` when I try to compile:
```java
@AvroName("Wine")
public enum WineWithAnnotations {
    Malbec,
    Shiraz,
    CabSav,
    Merlot
}
```

sksamuel:
Sorry I meant the AvroName provided by avro4s not the one that comes with
Avro itself.

stevewampler:
I tried that (using the Avro4s `@AvroName`) annotation, but it doesn't work because the `@AvroName` annotation is defined as a scala annotation and not a java annotation.
For example:
```scala
@AvroName("Wine")
public enum WineWithAnnotations {
    Malbec,
    Shiraz,
    CabSav,
    Merlot
}
```
Fails to compile:
```scala
Error:(5, 2) java: incompatible types: com.sksamuel.avro4s.AvroName cannot be converted to java.lang.annotation.Annotation
```
Any idea how to work around that?

sksamuel:
No, there may just be no way to do it. In which case I think what you've done already makes sense and we can merge.

stevewampler:
Thanks @sksamuel !

Issue #422, opened 2020-01-03
Performance of sealed traits and type parameters


andreas-schroeder:
I have been working with avro4s since 1,5 years and I must first say thank you for providing this very convenient library. It is really well documented all around and nice to use; thank you for the effort you put into it!

Recently however, I have experienced a significant performance impact of avro4s on the overall CPU utilization of a codebase. I managed to narrow it down to the performance impact of using sealed traits with one subtype having a type parameter. 

I've provided you a benchmark for encoding / decoding that shows the impact of sealed traits and type parameters, and compares as well with the performance of partially hand-rolling encoders and decoders. I will share the benchmarks as PR soon.

The numbers are roughly as follows:
```
::Benchmark avro4s simple field decoding::                           0.003976 ms
::Benchmark avro4s type union decoding::                             0.010974 ms
::Benchmark avro4s type parameter decoding::                         0.003747 ms
::Benchmark avro4s union type with type param decoding::             0.012863 ms
::Benchmark Avro specific record union type field decoding::         0.001309 ms
::Benchmark avro4s union type with type param hand-rolled decoding:: 0.00148  ms

::Benchmark avro4s simple field encoding::                           0.003463 ms
::Benchmark avro4s type union encoding::                             0.014125 ms
::Benchmark avro4s type parameter encoding::                         0.004532 ms
::Benchmark avro4s union type with type param encoding::             0.019654 ms
::Benchmark Avro specific record union type field encoding::         0.00087  ms
::Benchmark avro4s union type with type param hand-rolled encoding:: 0.002641 ms
```

So introducing a sealed trait (with 3 subtypes) makes encoding x 4.0 slower and decoding x 2.7 slower. Introducing a type parameter on its own has little impact, the combination however makes encoding x 5.6 slower and decoding x 3.2 slower in the benchmark.

Comparing this with a partially hand-rolled encoding / decoding is also interesting: encoding is 7.4 x slower than hand-rolled, and 22.6 x slower than specific avro record, and decoding is 8.7 x slower than hand-rolled, and 9.8 x slower than specific avro record. In my real use-case, the improvement of partially hand-rolling the encoding gives a ~ x 12 speedup.

I have been looking around the code (without deep knowledge of it) and pointed a profiler at it (without deep knowledge of it), and gained the impression that there is a lot of potential to move significant amount of logic from encoding / decoding paths to encoder / decoder generation.

[...]

PR #424, opened 2020-01-04
Improve encoding / decoding performance by caching

andreas-schroeder:
NameExtractors are expensive to build; this caches the instances.
Also, SchemaHelper.extractTraitSubschema compiles the same regex on each
call, this is now extracted.

Relates to issue #422


andreas-schroeder:
travis failed due to dependency download failure.

andreas-schroeder:
I continued to experiment with improvements, and it is much simpler and cleaner to do the caching in the Encoder / Decoders themselves. PR will follow up.

PR #425, opened 2020-01-04
Improve encoding / decoding performance

andreas-schroeder:
NameExtractors are expensive to build; this pre-computes the
names and fullnames needed outside of encoding / decoding.
Also, SchemaHelper.extractTraitSubschema compiles the same regex on each
call, this is now extracted.
Also, removed unused code.

Benchmark results:
```
::Benchmark avro4s union type with type param encoding:: 0.005265 ms
::Benchmark avro4s union type with type param decoding:: 0.00267 ms
```

Encoding speedup is x 2.6, decoding speedup is x 4.8.

Flame graphs.

Encoding before:
Encode before refactoring

Encoding after:
Encode after refactoring

Decoding before:
Decode before refactoring

Decoding after:
Decode after refactoring


andreas-schroeder:
Hi @sksamuel, just to inform you - the full rewrite for performance optimization is quite a huge effort, it might be very worthwhile to integrate these quick wins here. The other improvements are several weeks away I think.

sksamuel:
Agreed, merged.

PR #426, opened 2020-01-05
Draft implementation of codecs to allow for pre-computing schemas and mappings

andreas-schroeder:
This sketch is intended to show how schema computations and name mappings could be moved out of the critical encode / decode path. 

I tried to find the most challenging parts and address them, namely:
- how to handle sealed traits and case classes
- how to handle annotations on fields
- how to propagate annotations to primitive codecs.
Not sure though these actually are the most challenging parts.

For the design I decided to factor on the shape of the data that is encoded / decoded rather than to split schema generation, encoding, and decoding. I tried this since I feel the coupling is stronger this way. I see that having a SchemaFor[T] will be necessary in order to support custom encoding / decoding, and have build e.g. RecordCodec so that a SchemaFor could re-use the same code (RecordCodec.buildSchema).

A key piece on how to deal with annotations-based schema customization is the method Codec.withSchema. It allows to enrich a schema that has been derived bottom-up by Magnolia with top-down information available through field annotations.

I must say I'm not sure I fully understood the namespace-conversions that happen in the schema helpers encoding / decoding of sealed traits and case classes (i.e. GenericEncoder.encode and  SchemaHelper.extractTraitSubschema and the like), so while I think it could work, it still needs a test to actually prove it is working properly. On a second thought, I think TypeUnionCodec isn't working fully as before, as it doesn't propagate annotations of the sealed trait to the subtypes; this can be fixed, however.

I moved the FieldMapper to an implicit of the Magnolia-based codec derivation. That has the downside that it makes an implicitNotFound annotation on the Codec really necessary, as it is virtually impossible for a newcomer to understand what's going on if there is no implicit FieldMapper in scope.

Let me know what you think, and if you think it is worthwhile to proceed with this.


andreas-schroeder:
.. relates of course to issue #422 :)

andreas-schroeder:
Benchmark results:
```
::Benchmark avro4s union type with type param alternative codec encoding:: 0.001365 ms
::Benchmark avro4s union type with type param alternative codec decoding:: 0.00178 ms
```

Encoding speedup is x 9.4, decoding speedup is x 11.0

Encoding flamegraph:
Encoding

Decoding flamegraph:
Decoding

sksamuel:
This is amazing stuff. I haven't had chance to look through it entirely yet, but I will this week. I will look to do a release with the improvements once it's all merged and finalized.
And perhaps you could do a blog post, you should get the credit for this work.

andreas-schroeder:
Glad to hear that you are favorable towards this contribution, also thank you for your kind suggestion on how to get credit for this work, I will give it a thought! 

For proceeding, I have a question concerning the design: would it be better to stick with Encoder / Decoder or is it fine to switch to Codec being both (and by this having a guaranteed reference to the same Schema)? It could also be possible to introduce projections of Codec to Encoder / Decoder, but how to manage map and comap implementation of Encoder / Decoder could be tricky, as the origin Codec from which Encoder / Decoder would be derived would only support an inmap operation (invariant map). What do you think?

sksamuel:
> For proceeding, I have a question concerning the design: would it be better to stick with Encoder / Decoder or is it fine to switch to Codec being both (and by this having a guaranteed reference to the same Schema)? It could also be possible to introduce projections of Codec to Encoder / Decoder, but how to manage map and comap implementation of Encoder / Decoder could be tricky, as the origin Codec from which Encoder / Decoder would be derived would only support an inmap operation (invariant map). What do you think?

The invariance of map could be an issue. Other than that, we could look at bringing them together. I have seen advantages elsewhere with this, but also I wonder - why doesn't circe do it? Maybe because it's schemaless ?.

andreas-schroeder:
I found a solution by introducing a Functor and Cofunctor poor man's typeclass for Encoder and Decoder, I'll push a commit in a couple hours to show you.

andreas-schroeder:
@sksamuel have a look at `Encoder2`, `Decoder2` and the companion object of `Codec` for the definitions of map / comap / bimap.

andreas-schroeder:
Hi @sksamuel, so this turns out to be an even bigger change than I expected. Currently, not everything is fully settled down and final, but converging. A couple further changes will be needed to align everything.

My strategy was to create V2 versions and gradually migrate tests over from the original version to V2. Later in this branch, we could delete the original version and rename the V2's to the proper names.

I've got four things to clarify (search for "TODO clarify" in the changed files) of things where I wasn't sure if they should be implemented.

Additionally, I was surprised that the input stream allowed to define an intermediate schema that may diverge from the decoder's expectations. I decided to use the decoder's schema as the reader schema of the input stream - do you think that this covers all use cases?

Finally, after laying everything out, it might be actually better to drop the composite codecs as they introduce a lot of code duplications. Have a look and tell me what you think (especially on the record and type union encodings)...

andreas-schroeder:
Here are flamegraphs of decoding. With the refactoring, avro4s makes up 12 % of total time, and the majority of decoding time is spent in org.apache.avro code:
Decoder Ratio

The breakdown of avro4s decoding looks well evenly distributed:
Decoder Details

andreas-schroeder:
For encoding, avro4s interestingly still makes up 38% of total time:
Encoder Ratio

Although there are no obvious places where time is spent:
Encoder Details

andreas-schroeder:
Hi @sksamuel , I hope you are well. I've now completed the changes on this branch, and would love to get some feedback on these changes. 

Also, I would love to know how we should proceed towards getting this released. Especially, how should I approach resolving the conflicts with master - should I rebase this branch on top of maser (if so, I'd rather squash all my commits into one), or should I rather merge master into the branch.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroBinaryInputStream.scala
```diff
@@ -17,12 +17,10 @@ import scala.util.Try
   * input stream requires an instance of Decoder.
   */
 class AvroBinaryInputStream[T](in: InputStream,
-                               writerSchema: Schema,
-                               readerSchema: Schema,
-                               fieldMapper: FieldMapper = DefaultFieldMapper)
```
andreas-schroeder: Note: reader schema and field mapper are provided by the Decoder.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroDataInputStream.scala
```diff
@@ -10,17 +10,13 @@ import org.apache.avro.io.DatumReader
 import scala.util.Try
 
 class AvroDataInputStream[T](in: InputStream,
-                             writerSchema: Option[Schema],
-                             readerSchema: Option[Schema],
-                             fieldMapper: FieldMapper = DefaultFieldMapper)
```
andreas-schroeder: Same here: schema and mapper are provided by Decoder

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroDataOutputStream.scala
```diff
@@ -13,30 +13,27 @@ import org.apache.avro.generic.{GenericDatumWriter, GenericRecord}
   * case the [[AvroBinaryInputStream]] is what you would need.
   *
   * @param os      the underlying stream that data will be written to.
-  * @param schema  the schema that will be used to encode the data, sometimes called the writer schema
   * @param codec   compression codec
   * @param encoder the avro4s [[Encoder]] that will convert each value to a GenericRecord.
   */
 case class AvroDataOutputStream[T](os: OutputStream,
-                                   schema: Schema,
```
andreas-schroeder: schema and fieldmapper are provided by Encoder

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroInputStream.scala
```diff
@@ -68,19 +68,9 @@ class AvroInputStreamBuilderWithSource[T: Decoder](format: AvroFormat, in: Input
...
-  }
-
-  /**
-    * Builds an [[AvroInputStream]] with the specified reader and
-    * write schemas.
-    */
-  def build(writerSchema: Schema, readerSchema: Schema) = format match {
-    case DataFormat => new AvroDataInputStream[T](in, Some(writerSchema), Some(readerSchema))
-    case BinaryFormat => new AvroBinaryInputStream[T](in, writerSchema, readerSchema)
-    case JsonFormat => new AvroJsonInputStream[T](in, writerSchema, readerSchema)
```
andreas-schroeder: Since the reader schema is provided by Decoder, this method becomes obsolete

avro4s-core/src/main/scala/com/sksamuel/avro4s/ImmutableRecord.scala
```diff
@@ -10,7 +10,7 @@ import org.apache.avro.specific.SpecificRecord
   */
 trait Record extends GenericRecord with SpecificRecord
 
-case class ImmutableRecord(schema: Schema, values: Vector[AnyRef]) extends Record {
```
andreas-schroeder: Widening the type allows using an array on record creation which is quite efficient

avro4s-core/src/main/scala/com/sksamuel/avro4s/RecordFormat.scala
```diff
@@ -14,11 +13,10 @@ trait RecordFormat[T] extends ToRecord[T] with FromRecord[T] with Serializable
   */
 object RecordFormat {
 
-  def apply[T: Encoder : Decoder : SchemaFor]: RecordFormat[T] = apply(AvroSchema[T])
```
andreas-schroeder: No need for a separate SchemaFor, as Encoder and Decoder both provide a SchemaFor

avro4s-core/src/test/scala/com/sksamuel/avro4s/github/Github415.scala
```diff
@@ -3,25 +3,28 @@ package com.sksamuel.avro4s.github
...
-      case class Metrics(count: Int)
-      case class EarlyLate(early: Metrics)
-      case class Stats(session: Option[EarlyLate])
-    }
-
-    case class Rebuffers(network: Option[Rebuffers.Stats])
-
-    case class PlaybackSession(rebuffers: Option[Rebuffers])
-
+    // TODO clarify is it necessary to be able to serialize encoders / decoders of unserializable classes?
```
andreas-schroeder: the Rebuffers classes aren't serializable as it they are inside a non-serializable class (Github415). For the test I moved them therefore to the companion object. Why this was relevant is something I don't remember fully, but iirc this was due to the codecs referring to the Magnolia case classes that referenced the class.

andreas-schroeder:
superseded by #470

PR #429, opened 2020-01-08
Improve AvroBinaryInputStream by using GenericData singleton

lvitaly:
Improve AvroBinaryInputStream by using GenericDatumReader constructor with GenericData singleton instead of creating new instances of GenericData.

https://github.com/apache/avro/blob/4cd778a7e0de048eacf2a06abe3d0e310bbcc343/lang/java/avro/src/main/java/org/apache/avro/generic/GenericDatumReader.java#L61


sksamuel:
Nice spot.

PR #437, opened 2020-02-10
Fix recursive schema

### LINKED ISSUE #250: Recusrive schemas don't work in latest version
Trying the example in the README https://github.com/sksamuel/avro4s#recursive-schemas and it doesn't compile.

dragisak:
Fixes #250 

While traversing the graph, cache schemas created so far.

Unfortunately, I had to make a API change to:
```scala
trait SchemaFor[T] {
    def schema(fieldMapper: FieldMapper, context: SchemaFor.Context): Schema
}
```

I am open to any suggestion on how to avoid exposing internal implementation like this.


sksamuel:
Thanks for the PR.
Can you explain what the context is used for ?

dragisak:
To allow recursive schemas, you have to cache all the Schema's seen so far. When you traverse object graph, if Schema for your name/namespace pair has already been created, you can stop traversal and use that cached Schema. That's how avro [SchemaBuilder](https://avro.apache.org/docs/current/api/java/org/apache/avro/SchemaBuilder.FieldBuilder.html#type-java.lang.String-java.lang.String-) does it.

I had to figure a way to pass around that schema cache. Simplest was to add `Context` parameter to `schema` method.

sksamuel:
Where is SchemaFor.Context defined? I can't see that in your PR?

dragisak:
Here https://github.com/sksamuel/avro4s/pull/437/files#diff-8a05235c28d1810859da3b817e7441c7R578-R585

sksamuel:
Thanks for the replying @dragisak. I think github was doing that thing where it hides large diffs and I missed this.

So in magnolia, we have recursive support. Do you think it's better to use that, or pass the context around? We couldn't change the SchemaFor contract without bumping a major version.

WesselVS:
I also created a workaround for this much needed feature. I have a slightly different approach, where I leave SchemaFor intact and introduce SchemaForRec to generate recursive definitions. This is not ideal, but for now, it serves my needs. #453

sksamuel:
I feel like we should combine the approaches in #453 #426 #409 as a 4.0 release (allowing for breaking changes). What do you think ?

dragisak:
Sounds good. :+1

dragisak:
Closing this in favor of #453

Issue #441, opened 2020-02-22
Avro4s isn't generating unique schema names for deeply nested generic types


stevewampler:
I need avro4s to generate unique schema names for generic types nested more than one level.

For example, avro4s generates the truncated name `MatchEvent__Advertiser_Option` for the type `MatchEvent[Advertiser, Option[Seq[Organization]]]]`.

A unique name, that captures the actual type, would be `MatchEvent__Advertiser_Option_Seq_Organization`.

stevewampler:
Hi @sksamuel, would you please generate a 3.0.8 build? I'd like to use this fix in my project at work. Or, if there's a way for me to kick off a 3.0.8 build/deploy, let me know how and I'll do it.

sksamuel:
Releasing 3.0.8 now.

stevewampler:
Thanks @sksamuel !

sksamuel:
Should show up in a bit. It's been released.

PR #442, opened 2020-02-22
#441 Changed the implementation of the TypeInfo.fromTypeName method t…

stevewampler:
I changed the implementation of the `TypeInfo.fromTypeName` method to property generate `TypeInfo` hierarchies for nested generic types. Doing so fixes issue #441 where avro4s was truncating the schema names of deeply nested generic types, resulting in schema names that are not unique.

For example, before the fix, the type `A[B[C]]]` would result in a truncated schema name of `A__B`. With the fix, the name is now `A__B__C`, which correctly reflects the nested type. Likewise, if the type was `A[B[D]]]`, the resulting schema's name is now `A__B__D` and does not conflict with `A[B[C]]]`'s schema name of `A__B__C`.

Added two new schema tests to GenericSchemaTest to test deeply nested generic types. In addition to the new tests passing, all of the existing tests pass as well.


sksamuel:
Nice work

stevewampler:
Thanks @sksamuel ! Much appreciated.

Issue #447, opened 2020-03-01
GenericSerde.scala could also support JSON avro


sv3ndk:
Hi, 

Thanks for creating Avro4s, really helpful !

I recently wanted to use [GenericSerde.scala](https://github.com/sksamuel/avro4s/blob/master/avro4s-kafka/src/main/scala/com/sksamuel/avro4s/kafka/GenericSerde.scala), though needed support for JSON avro for easy debugging, so I ended up duplicating it :

https://github.com/sv3ndk/kafka-to-rdbms-export-akka-demo/blob/master/src/main/scala/com/svend/demo/Avro4sSerdes.scala

Would you like me to submit a PR back to Avro4S with this minor update? 

Also, the readme does not mention the existence of that class, which I find very useful, so I can add a few words about it there as well if you want?

sksamuel:
What did you change exactly ?

sv3ndk:
I added the optional constructor parameter `(jsonFormat: Boolean = false)` that I then use to select the Avro format:  `if (jsonFormat) AvroInputStream.json[T] else AvroInputStream.binary[T]` and `if (jsonFormat) AvroOutputStream.json[T] else AvroOutputStream.binary[T]`

sksamuel:
I think that's good but we could use the AvroFormat trait and then support all 3 formats ?

sv3ndk:
Hi, 

Sorry for the week of silence, I'm back now :)

Good idea to rely on `AvroFormat`. I gave it a try and just submitted a MR here: https://github.com/sksamuel/avro4s/pull/452.

I suggest we continue the discussion over there?

PR #452, opened 2020-03-08
Generic serdes

sv3ndk:
As discussed in https://github.com/sksamuel/avro4s/issues/447

This MR adds support for Json and Data Avro format in the Kafka `GenericSerde` class and documents its usage `readme.md`. By default, the behaviour is the same as before, i.e usage of `BinaryFormat`.

Behaviour is validated with a few supplementary UT, and I did a quick manual validation with Akka Streams (https://github.com/sv3ndk/kafka-to-rdbms-export-akka-demo/blob/master/src/main/scala/com/svend/demo/DataGenerator.scala#L27).

### DIFF
--- a/README.md
+++ b/README.md
@@ -916,6 +916,27 @@ val format = RecordFormat[Composer]
 val ennio = format.from(record)
 ```
 
+## Usage as a Kafka Serde
+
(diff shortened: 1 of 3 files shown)


sksamuel:
Looks great.

PR #455, opened 2020-03-12
Fix decoder for coproducts with arrays/maps

peterneyens:
After upgrading to the latest `mu-scala` release, we also are using avro4s 3.x.
I had some problems with decoding some union types where one of the elements is an array.

It seems we currently are attempting to decode the elements of the union with specific decoders but still with the schema of the complete union.

Here I unfold the schemas for union types and filter to the schemas that make sense for the `SafeFrom` instance that we are building. Then we try them one by one until one of them succeeds, otherwise we throw an error which includes the error messages of all the schemas we have tried.

Today was the first time looking under the cover of `avro4s` so I am not sure if these changes are within the right spirit? I can probably try to clean this up a bit if we think the solution looks OK.


sksamuel:
:+1:

fedefernandez:
hey @sksamuel, do you have plans for releasing a new version with these changes? Thanks

PR #456, opened 2020-03-12
Fix nested coproducts

peterneyens:
This feels dirty, but adding a special case for `Coproduct` in `SafeFrom` fixes decoders for nested coproducts.

Maybe we should put this in the `:+:` decoder?


juanpedromoreno:
Hi @sksamuel, 

Do you have any plans for releasing a new version with these changes? Thanks!

sksamuel:
3.0.10 is released.

PR #468, opened 2020-04-20
don't pull in scala-compiler as run-time dependency

mberndt123:
Hi,

I recently noticed that when you use avro4s, you end up with scala-compiler on the run-time classpath. But it's needed only at compile-time, and including it in the run-time classpath adds unnecessary bloat for the users.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -15,7 +15,7 @@ object Build extends AutoPlugin {
     val CatsVersion = "2.0.0-RC2"
     val ShapelessVersion = "2.3.3"
     val RefinedVersion = "0.9.13"
-    val MagnoliaVersion = "0.14.5"
+    val MagnoliaVersion = "0.16.0"
     val ScalaMeterVersion = "0.19"
   }
 


mberndt123:
Unfortunately this change is not sufficient to keep scala-compiler out of the run-time classpath, because magnolia depends on mercator, and mercator still pulls it in.

I've fixed that here:
https://github.com/propensive/mercator/pull/20
When a new version with that fix is released, a new version of magnolia needs to be released and avro4s needs to upgrade to that, and then this should be fixed. *sigh*

sksamuel:
How big is the magnolia dependency ?

mberndt123:
Hey @sksamuel,

I made a mistake when I reported the bug. Magnolia is in fact needed at run-time, but scala-compiler is not. The scala-compiler jar is around 10 MB.

mberndt123:
@sksamuel,

I just learned that exclusions are also published in the POM file, so we don't need to wait for a new mercator/magnolia release to fix this. It's enough to exclude these dependencies in the avro4s build.sbt, so I've added that change as well.

Again, to clarify: unlike I thought initially, avro4s users won't need to update their build.sbt for this to work, they only need to bump their avro4s version to get rid of this unnecessary dependency.

mberndt123:
Hey @sksamuel,

is there any chance we can get this one merged?

mberndt123:
Closing this because there's a new Magnolia version that doesn't have the problematic dependency (and because @sksamuel doesn't seem to care)

sksamuel:
@sksamuel has been busy with other others and you can easily fork this and release your own version. That's why open source is so good, you don't need the original author to do everything for you.

mberndt123:
Oh, so you do care? Great, so you can merge it now, right?

mberndt123:
@sksamuel:
Oh, so you do care? Great, so you can merge it now, right?

mberndt123:
Thanks for merging @sksamuel !

sksamuel:
You're welcome. The snapshot will be built automatically if you want to
test it out.

PR #460, opened 2020-04-06
Pre-compute name mappings and schemas in encoder/decoders

andreas-schroeder:
Followup to #426 - rebased it on top of master.

PR #470, opened 2020-04-27
Support recursive schemas while improving performance

### LINKED ISSUE #250: Recusrive schemas don't work in latest version

andreas-schroeder:
This is a followup on #460 (which itself was a followup on #426) and takes the ideas from @WesselVS and applies it to the avro4s variant with precomputed mappings and Schemas.

This had some interesting consequences, especially I needed to completely drop `Codec`s because it would have made the codebase too complex and it seems it isn't worth it.

The design I chose was based on two assumptions and is driven by constraints of Scala's implicit resolution that made the outcome somewhat unexpected:

* The first assumption for the design is that we want to keep using `Encoder` and `Decoder` traits for both atomic codecs (String, int, ...) and composite ones (records, type unions, list, maps) instead of introducing subtypes `AtomicEncoder` and `CompositeEncoder` (or the like).
* The second is that the codebase should be kept simple; in the new design with precomputed name mappings, a duplication of the implicit resolution in `Encoder` (for non-recursive schemas) and `EncoderRec` (for recursive ones) would actually make it necessary to maintain two codebases that follow a different approach for passing down schemas modifications (such as namespace overrides and Avro fixed schemas).

The limitation on implicit resolution that I experienced is that it detects a diverging resolution as soon as there are back and forth conversions between `EncoderRec` and `Encoder` (same applies for `SchemaFor` and `Decoder`, but I describe `Encoder` here in the following). Unfortunately, both conversions are necessary to a) use regular `Encoder`s in recursive ones and b) summon an `Encoder` no matter whether the `Encoder` is potentially composite (and therefore potentially recursive) or atomic.

So what I came up with in the end is to use only one type `Encoder` for implicit resolution and use that same type for composite (and potentially recursive) encoders. That means that the base trait `Encoder` features methods that are only needed because of composite encoders (which in this code, I dubbed `ResolvableEncoder` as they need some resolution/lookup mechanism to be ready to go.

Let me know if you think this is good to go, and where the code needs more comments or clearer structure. @WesselVS or @sksamuel, one of you could have a look and give me some feedback on whether you think this approach is viable? I feel I shouldn't proceed further alone with this without guidance from your side. I actually went pretty far ahead without keeping in sync with anyone else on the repo.

Fixes #250


andreas-schroeder:
also relates to issue #422

andreas-schroeder:
merged to `release/4.0.x`

WesselVS:
Hi Andreas, can't really find the time to look at your changes. Maybe later at some point. Wanted to say I'm really impressed and thankful for the stuff you've done! It's going to benefit my team a lot.

sksamuel:
I plan on going over this this weekend to give feedback.

andreas-schroeder:
Hi @sksamuel thanks for looking over the changes. For your review please consider that there are still rough edges in my proposal that I'm aware of and plan to fix in the next days. More specifically, I plan to:
* Get rid of the Resolvable supertype - it' not really useful and complicates things (it prevents defining the Codec subtype of Encoder and Decoder).
* Rename the resolution methods from apply() to a more meaningful name (resolveDecoder, e.g.).
* Get the implicit method names an parameter names consistent. Some still have a -Rec suffix.
* Use proper visibility and make the resolution methods final on Encoder/Decoder/SchemaFor that aren't for overriding.
* Fix the implicit resolution conflicts that prevented grouping all implicits into traits.
So if one of these bother you, know that they bother me too and I'll fix them 😊

PR #478, opened 2020-04-30
Pre-compute field name indices to reduce access time in ImmutableRecord.get

andreas-schroeder:
Improvement suggestion based on #392 and #385

@sksamuel, @yeryomenkom wdyt?


andreas-schroeder:
Answering to myself why this is a bad idea: it would spend time for verifying the assertions and building the index map in vain if the access was via index anyway. We'd have to thoroughly measure this to validate it actually improves runtime.

andreas-schroeder:
I think the best approach would be to implement it as in Avro GenericData: https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/generic/GenericData.java#L252-L257

sksamuel:
#491 supersedes this ?

andreas-schroeder:
It's not superseded by the 4.0.x, as it's not yet contained in there. As this avoid a) converting the schema field collection to a Scala collection and b) a linear scan through the fields, I'd suggest to include it in 4.0.x

PR #491, opened 2020-05-18
Release 4.0.0

andreas-schroeder:
Release 4.0.0 intents to 
* improve performance of record encoding and decoding, and 
* add back support for recursive types.


avro4s-core/src/main/scala/com/sksamuel/avro4s/DefinitionEnvironment.scala
```diff
@@ -0,0 +1,29 @@
...
+import scala.reflect.runtime.universe._
+
+/**
+ * Helps with encoders/decoders/schemas for recursive types. Provides a map of already defined bindings for types.
+ * For an Encoder e.g., captures which encoders are already defined when building the (cyclic) encoder graph for
+ * a recursive type.
+ *
+ * @tparam Typeclass Encoder, Decoder or SchemaFor
+ */
+class DefinitionEnvironment[Typeclass[_]](
```
sksamuel, line 12: DefinitionEnvironment is used keeps a map of typess we have already seen and the typeclasses returned for those?Magnolia is meant to handle recursive types for us, does this fail ?

avro4s-core/src/main/scala/com/sksamuel/avro4s/RecordFormat.scala
```diff
@@ -14,11 +13,10 @@ trait RecordFormat[T] extends ToRecord[T] with FromRecord[T] with Serializable
   */
 object RecordFormat {
 
-  def apply[T: Encoder : Decoder : SchemaFor]: RecordFormat[T] = apply(AvroSchema[T])
+  def apply[T: Encoder : Decoder]: RecordFormat[T] = new RecordFormat[T] {
+    private val fromRecord = FromRecord[T]
+    private val toRecord = ToRecord[T]
 
-  def apply[T: Encoder : Decoder](schema: Schema)(implicit fieldMapper: FieldMapper = DefaultFieldMapper): RecordFormat[T] = new RecordFormat[T] {
```
sksamuel, line 19: Was the schema just not being used inside the ToRecord and FromRecord ?

avro4s-core/src/main/scala/com/sksamuel/avro4s/DatatypeShape.scala
```diff
@@ -0,0 +1,34 @@
+package com.sksamuel.avro4s
+
+import magnolia.{CaseClass, SealedTrait}
+
+import scala.reflect.runtime.universe._
+
+object DatatypeShape {
+  def of[TC[_], T: WeakTypeTag](ctx: SealedTrait[TC, T]): SealedTraitShape = {
```
sksamuel, line 8: the ctx param isn't used.

avro4s-core/src/main/scala/com/sksamuel/avro4s/MagnoliaDerived.scala
```diff
@@ -0,0 +1,89 @@
+package com.sksamuel.avro4s
+
+import magnolia._
+
+import scala.language.experimental.macros
+import scala.reflect.runtime.universe._
+
+trait MagnoliaDerivedSchemaFors {
```
sksamuel, line 8: I really like the way you tidied this up using DatatypeShapes

avro4s-core/src/main/scala/com/sksamuel/avro4s/MagnoliaDerived.scala
```diff
@@ -0,0 +1,89 @@
...
+      env.get[T].getOrElse {
+        DatatypeShape.of(ctx) match {
+          case SealedTraitShape.TypeUnion => TypeUnions.schema(ctx, env, update)
+          case SealedTraitShape.ScalaEnum => SchemaFor[T](ScalaEnums.schema(ctx), DefaultFieldMapper)
+        }
+      }
+    }
+  }
+
+  def combine[T: WeakTypeTag](ctx: CaseClass[SchemaFor, T])(implicit fieldMapper: FieldMapper = DefaultFieldMapper): SchemaFor[T] = {
```
sksamuel, line 24: I'm trying to understand why the field mapper was moved out of Encoder / Decoder into here. I guess its because it's only needed when generating certain encoders rather than all the time. Can you clarify for me?

avro4s-core/src/main/scala/com/sksamuel/avro4s/Codec.scala
```diff
@@ -0,0 +1,38 @@
+package com.sksamuel.avro4s
+
+
+trait Codec[T] extends Encoder[T] with Decoder[T] with SchemaAware[Codec, T] { self =>
```
sksamuel, line 4: Can you add some scaladoc or an explanation here of what codecs do exactly ?

avro4s-core/src/main/scala/com/sksamuel/avro4s/DefinitionEnvironment.scala
andreas-schroeder: One complication I see when using Magnolia's capabilities for recursive encoders and decoders is that the caching of name mappings would need to introduce synchronization locks to be correct.

sksamuel: Thank you that's perfect. Would you mind adding what you wrote as documentation onto the def environment so it's saved for future readers.

avro4s-core/src/main/scala/com/sksamuel/avro4s/TypeUnions.scala
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  private def enrichedSubtypes[Typeclass[_], T](ctx: SealedTrait[Typeclass, T],
+                                                update: SchemaUpdate): Seq[(Subtype[Typeclass, T], SchemaUpdate)] = {
+    val enrichedUpdate = update match {
+      case NoUpdate =>
+        // in case of namespace annotations, pass the namespace update down to all subtypes
+        val ns = new AnnotationExtractors(ctx.annotations).namespace
+        ns.fold[SchemaUpdate](NoUpdate)(NamespaceUpdate)
+      case _ => update
```
sksamuel, line 95: Would the annotated namespace be missed here since if we have a namespace/full update, we don't inspect the trait? I guess this means we're just assuming you annotate the trait itself, which I think is fine.

avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaUpdate.scala
```diff
@@ -0,0 +1,9 @@
+package com.sksamuel.avro4s
+
+sealed trait SchemaUpdate
+
+object SchemaUpdate {
```
sksamuel, line 5: I'd appreciate some docs on this as to why it exists as an ADT rather than a method.
I gather it's being used as a command to update a schema by replacing the namespace or replacing the entire thing (for when we resolve a schema?). What was the rational for not just having some helper method.

sksamuel:
So overall this is a huge improvement on what I had before. The code is elegant and concise and easy to read, I'm very pleased with the result. I think that it's ready to merge.

avro4s-core/src/main/scala/com/sksamuel/avro4s/TypeUnions.scala
andreas-schroeder: The code gives priority to explicit schema customizations (via `withSchema`) over annotations on the trait. If the trait isn't annotated but some of the elements are, these namespace annotations will be handled in `Records`.

avro4s-core/src/main/scala/com/sksamuel/avro4s/SchemaUpdate.scala
andreas-schroeder: I will add some docs. Replacing the entire schema happens when the client uses the Encoder / Decoder's `withSchema` method to override the inferred schema with a custom one. The client-provided schema is then decomposed along the magnolia-inferred ADT structure to create updates for field encoders / decoders, using a similar path as for namespace updates. So basically, this ADT exists because namespace updates and full schema updates travel in the same direction (from traits to case classes, and from fields to case classes).

andreas-schroeder:
@sksamuel thanks for your review and approval; I am really thankful for all the tests you to care to collect and maintain. This big effort wouldn't have been possible without them!

sksamuel:
There's a couple of conflicts that I can't edit, so once they're ready, please merge!

andreas-schroeder:
Merged to master with conflicts resolved.

PR #494, opened 2020-06-01
Added encoder builder

sksamuel:
Just an idea on something we could do. It doesn't use anything clever, and isn't super terse, but you don't need any FP knowledge to use it.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/Encoder.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/Encoder.scala
@@ -1,11 +1,14 @@
 package com.sksamuel.avro4s
 
 import com.sksamuel.avro4s.Encoder.DelegatingEncoder
+import com.sksamuel.avro4s.EncoderHelpers.FieldEncoder
 import com.sksamuel.avro4s.SchemaUpdate.{FullSchemaUpdate, NoUpdate}
 import org.apache.avro.{Schema, SchemaBuilder}
 import org.apache.avro.generic.{GenericData, GenericRecord}
 
+import scala.reflect.runtime.universe.WeakTypeTag
 import scala.language.experimental.macros
+import scala.collection.JavaConverters._
 
 /**
   * An [[Encoder]] encodes a Scala value of type T into a compatible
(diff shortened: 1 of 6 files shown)


avro4s-core/src/main/scala/com/sksamuel/avro4s/Encoder.scala
```diff
@@ -121,6 +122,45 @@ object Encoder
     }
   }
 
+  case class EncoderField[T](name: String, schema: Schema, extractor: T => Any)
+
+  object EncoderField {
+    def string[T](name: String, extractor: T => Any) = EncoderField(name, Schema.create(Schema.Type.STRING), extractor)
```
andreas-schroeder, line 128: shouldn't it be
```scala
def field[T, F](name: String, extractor: T => F)(implicit encoder: Encoder[F]): EncoderField[T, F] = ...
```
?

avro4s-core/src/main/scala/com/sksamuel/avro4s/Encoder.scala
```diff
@@ -121,6 +122,45 @@ object Encoder
...
+    val recordSchema = fields.foldLeft(builder.fields()) { case (acc, field) =>
+      acc.name(field.name).`type`(field.schema).noDefault()
+    }.endRecord()
+
+    new Encoder[T] {
+
+      override def encode(value: T): AnyRef = {
+        val record = new GenericData.Record(recordSchema)
+        for (field <- fields) {
+          record.put(field.name, field.extractor(value))
```
andreas-schroeder, line 155: then here something like

```scala
record.put(field.name, field.encoder.encode(field.extractor(value)))
```

avro4s-core/src/main/scala/com/sksamuel/avro4s/Encoder.scala
andreas-schroeder: this would make it possible to use the schema from the `encoder` too

andreas-schroeder:
yep, that's pretty terse and with the use of implicit encoders, that's pretty much it for the encoding side. On the decoding side though ... things will be a bit more involved.

avro4s-core/src/main/scala/com/sksamuel/avro4s/Encoder.scala
sksamuel: Done.

avro4s-core/src/main/scala/com/sksamuel/avro4s/Encoder.scala
sksamuel: Done.

andreas-schroeder:
@sksamuel I was thinking more like this: https://github.com/andreas-schroeder/avro4s/commit/899fa35fb51a9aad1deacc71f026985cc3b59fda (the name `field` is kind of redundant, tho)

andreas-schroeder:
Just thought of one thing: we should  implement ResolvableEncoder to make it play along with recursive types.

andreas-schroeder:
I've implemented the resolvable variant, but I couldn't get it picked up properly by implicit search on recursive types. In the field list expression, the definition of the Encoder is not yet available, so that implicit search picks Encoder.gen instead. Using def instead of val doesn't change things, unfortunately.

andreas-schroeder:
Nevermind, here it is: https://github.com/andreas-schroeder/avro4s/commit/c101bff6fadb291778b212687c581b0d83b3a99a it's actually necessary to add a type annotation to the implicit def for it to work. What do you think?

andreas-schroeder:
By the way, feel free to refactor / reorganize my suggestion, I saw that you prefer to put classes top level, while I put my addition into the companion object.

andreas-schroeder:
Fun with Macros for the Decoder (currently recursion-unaware): https://github.com/andreas-schroeder/avro4s/commit/42a94633fa74c3558156a200ec42202842b14d84
Downside is that default parameter values are not supported in macros, so doc, props, and aliases must be either given explicitly as empty, or we need to find another way to define the basic record schema info.

andreas-schroeder:
Here's how to support default values (approach copied from scala logging): https://github.com/andreas-schroeder/avro4s/commit/6f26438ca76d01d3d2e23cb6faf0b9056c55c3f4

andreas-schroeder:
I've added support for recursive types in decoder (https://github.com/andreas-schroeder/avro4s/commit/6cba516483815895bee38e203260d99bdb6d81f8) and we can also allow fully customizing the extraction of a field value from a GenericRecord (https://github.com/andreas-schroeder/avro4s/commit/148b1ba041c6e0e76b1bb810db39c31c74f98c78)

I found some simplifications we could also apply to the encoder side.

PR #498, opened 2020-06-12
Benchmarking direct binary encoding / decoding

andreas-schroeder:
```
[info] Benchmark                     Mode  Cnt        Score        Error  Units
[info] Decoding.avro4sGenerated     thrpt   10  4009718,475 ±  69735,904  ops/s
[info] Decoding.avro4sHandrolled    thrpt   10  4861150,336 ±  41761,760  ops/s
[info] Decoding.avroSpecificRecord  thrpt   10  3997881,450 ± 174337,201  ops/s
[info] Decoding.binaryDecoder       thrpt   10  7758766,011 ± 106929,923  ops/s
[info] Encoding.avro4sGenerated     thrpt   10  3659541,561 ±  52548,236  ops/s
[info] Encoding.avro4sHandrolled    thrpt   10  4781688,322 ±  52322,058  ops/s
[info] Encoding.avroSpecificRecord  thrpt   10  5485834,367 ± 115538,193  ops/s
[info] Encoding.binaryEncoder       thrpt   10  9824665,634 ± 161617,669  ops/s
```

### DIFF
--- a/benchmarks/src/main/scala/benchmarks/Decoding.scala
+++ b/benchmarks/src/main/scala/benchmarks/Decoding.scala
@@ -21,6 +21,8 @@ object Decoding extends BenchmarkHelpers {
       new RecordWithUnionAndTypeField(new ValidInt(255, t)).toByteBuffer
     }
 
+    val avro4sSchema = AvroSchema[RecordWithUnionAndTypeField]
+
     val avro4sBytes = encode(RecordWithUnionAndTypeField(AttributeValue.Valid[Int](255, t)))
 
     val (handrolledDecoder, handrolledReader) = {
@@ -58,11 +60,10 @@ class Decoding extends CommonParams with BenchmarkHelpers {
 
   def decode[T](bytes: ByteBuffer, decoder: Decoder[T], reader: GenericDatumReader[GenericRecord]): T = {
     val dec =
-      DecoderFactory.get().binaryDecoder(new ByteBufferInputStream(Collections.singletonList(bytes.duplicate)), null)
+      DecoderFactory.get().directBinaryDecoder(new ByteBufferInputStream(Collections.singletonList(bytes.duplicate)), null)
     val record = reader.read(null, dec)
     decoder.decode(record)
   }
-
 
   @Benchmark
   def avroSpecificRecord(setup: Setup, blackhole: Blackhole) = {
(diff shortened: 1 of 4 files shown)


andreas-schroeder:
@sksamuel so this is what I had been talking about - the binary codec is a sketch that tries to mimic what the code would look like with Magnolia and some JIT optimization.

Here, I also fixed the decoding to use a DirectBinaryDecoder instead of one having a buffer.

In total, the numbers show that avro4s is in the same ballpark as specific record right now, and using binary decoder / encoder would roughly double the throughput.

sksamuel:
Awesome.

andreas-schroeder:
Well :D we have to think whether we want to support this and how to integrate this in the current codebase. The simple way would be to add methods to Encoder and Decoder. However, if a user wanted to only encode/decode GenericRecords and had custom Encoders and Decoders, she would need to provide the binary encoding/decoding methods even if they weren't needed.

The cleanest way would be to have a separate type, and then this could be a lot of code to duplicate and maintain.

mberndt123:
If you ask me, the existing Decoder stuff should be thrown out entirely. A signature like `def decode(value: Any): T` is insane.

mberndt123:
Hey @andreas-schroeder,

I'd like to discuss this code in `trait BinaryCodec`:
```
val map = ThreadLocal.withInitial[mutable.Map[Schema, ResolvingDecoder]](() => new mutable.HashMap())
```

The purpose appears to be to cache the `ResolvingDecoder`s so that a new one doesn't have to be created every time a value is read. I think a more elegant solution is to employ partial evaluation in the `readBytes` method like so:
```
  def readBytes(writerSchema: Schema): ByteBuffer => T = {
    val decoder = DecoderFactory.get.resolvingDecoder(Schema.applyAliases(writerSchema, schema), schema, null)

    { bytes =>
      val in = new ReusableByteBufferInputStream()
      in.setByteBuffer(bytes)
      val binary = DecoderFactory.get().directBinaryDecoder(in, null)
      decoder.configure(binary)
      read(decoder)
    }
  }
```

This way the user can apply the `readBytes` method to a `Schema` once, and it will return a function that can be called in a loop. No need for ThreadLocal mutable maps or other scary things.

andreas-schroeder:
Avro's `ResolvingDecoder` keep mutable state internally and therefore aren't thread-safe, and at the same time, they are really expensive to build, so should better be reused. In Avro's `GenericDatumReader`, there is the same `ThreadLocal`-based logic place to achieve this (it's turtles all the way down... even if wrapped nicely in functional abstractions). Running the benchmarks without replicating the logic that ensures thread-safety wouldn't be an even ground for comparison. Of course, the benchmarks could have been done the other way by explicitly passing a `ResolvingDecoder` to `GenericDatumReader`, and do the same in the `BinaryCodec`. I went for the other way since I expect that this will be the more intuitive external interface for users.

PR #499, opened 2020-06-12
small performance optimization

mberndt123:
Hey,

I've noticed that the code in Avro4s doesn't conform to the recommendations in the avro4s manual.
https://avro.apache.org/docs/1.9.2/gettingstartedjava.html
They recommend reusing the same `GenericRecord` over and over to avoid allocations.  Here is the relevant quote:
> Note how we perform the iteration: we create a single GenericRecord object which we store the current deserialized user in, and pass this record object to every call of dataFileReader.next. This is a performance optimization that allows the DataFileReader to reuse the same record object rather than allocating a new GenericRecord for every iteration, which can be very expensive in terms of object allocation and garbage collection if we deserialize a large data file.

I'm not sure how much of a difference that makes, but it can't hurt.

I also removed an `asInstanceOf` and an unnecessary field from the `AvroDataInputStream` class along the way.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroBinaryInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroBinaryInputStream.scala
@@ -24,8 +24,12 @@ class AvroBinaryInputStream[T](in: InputStream,
   private val avroDecoder = DecoderFactory.get().binaryDecoder(in, null)
 
   private val _iter = new Iterator[GenericRecord] {
+    var record: GenericRecord = null
     override def hasNext: Boolean = !avroDecoder.isEnd
-    override def next(): GenericRecord = datumReader.read(null, avroDecoder)
+    override def next(): GenericRecord = {
+      record = datumReader.read(record, avroDecoder)
+      record
+    }
   }
 
   /**
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroDataInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroDataInputStream.scala
@@ -4,8 +4,7 @@ import java.io.InputStream
 
 import org.apache.avro.Schema
 import org.apache.avro.file.DataFileStream
-import org.apache.avro.generic.{GenericData, GenericRecord}
-import org.apache.avro.io.DatumReader
+import org.apache.avro.generic.{GenericDatumReader, GenericRecord}
 
 import scala.util.Try
 
@@ -14,26 +13,26 @@ class AvroDataInputStream[T](in: InputStream,
                             (implicit decoder: Decoder[T]) extends AvroInputStream[T] {
 
 
-  // if no reader or writer schema is specified, then we create a reader that uses what's present in the files
-  private val datumReader = writerSchema match {
-    case Some(writer) => GenericData.get.createDatumReader(writer, decoder.schema)
-    case None => GenericData.get.createDatumReader(null, decoder.schema)
+  private val dataFileReader = {
+    // if no reader or writer schema is specified, then we create a reader that uses what's present in the files
+    val datumReader = new GenericDatumReader[GenericRecord](writerSchema.orNull, decoder.schema)
+    new DataFileStream[GenericRecord](in, datumReader)
   }
 
-  private val dataFileReader = new DataFileStream[GenericRecord](in, datumReader.asInstanceOf[DatumReader[GenericRecord]])
-
   override def iterator: Iterator[T] = new Iterator[T] {
+    var record: GenericRecord = null
     override def hasNext: Boolean = dataFileReader.hasNext
     override def next(): T = {
-      val record = dataFileReader.next
+      record = dataFileReader.next(record)
       decoder.decode(record)
     }
   }
 
   override def tryIterator: Iterator[Try[T]] = new Iterator[Try[T]] {
+    var record: GenericRecord = null
     override def hasNext: Boolean = dataFileReader.hasNext
     override def next(): Try[T] = Try {
-      val record = dataFileReader.next
+      record = dataFileReader.next(record)
       decoder.decode(record)
     }
   }
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroJsonInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroJsonInputStream.scala
@@ -15,8 +15,10 @@ final case class AvroJsonInputStream[T](in: InputStream,
   private val datumReader = new DefaultAwareDatumReader[GenericRecord](writerSchema, decoder.schema)
   private val jsonDecoder = DecoderFactory.get.jsonDecoder(writerSchema, in)
 
+  private var record: GenericRecord = null
   private def next = Try {
-    datumReader.read(null, jsonDecoder)
+    record = datumReader.read(record, jsonDecoder)
+    record
   }
 
   def iterator: Iterator[T] = Iterator.continually(next)
--- a/benchmarks/src/main/scala/benchmarks/Decoding.scala
+++ b/benchmarks/src/main/scala/benchmarks/Decoding.scala
@@ -1,11 +1,12 @@
 package benchmarks
 
-import java.io.ByteArrayOutputStream
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
 import java.nio.ByteBuffer
 import java.util.Collections
 
 import benchmarks.record._
 import com.sksamuel.avro4s._
+import org.apache.avro.file.CodecFactory
 import org.apache.avro.generic.{GenericDatumReader, GenericDatumWriter, GenericRecord}
 import org.apache.avro.io.{DecoderFactory, EncoderFactory}
 import org.apache.avro.util.ByteBufferInputStream
@@ -13,16 +14,7 @@ import org.openjdk.jmh.annotations._
 import org.openjdk.jmh.infra.Blackhole
 
 object Decoding extends BenchmarkHelpers {
-  @State(Scope.Thread)
-  class Setup {
-    val avroBytes = {
-      import benchmarks.record.generated.AttributeValue._
-      import benchmarks.record.generated._
-      new RecordWithUnionAndTypeField(new ValidInt(255, t)).toByteBuffer
-    }
-
-    val avro4sBytes = encode(RecordWithUnionAndTypeField(AttributeValue.Valid[Int](255, t)))
-
+  trait Setup {
     val (handrolledDecoder, handrolledReader) = {
       import benchmarks.handrolled_codecs._
       implicit val codec: Codec[AttributeValue[Int]] = AttributeValueCodec[Int]
@@ -38,6 +30,33 @@ object Decoding extends BenchmarkHelpers {
       val reader = new GenericDatumReader[GenericRecord](decoder.schema)
       (decoder, reader)
     }
+  }
+
+  @State(Scope.Thread)
+  class AvroBytesSetup extends Setup {
+    val avroBytes = {
+      import benchmarks.record.generated.AttributeValue._
+      import benchmarks.record.generated._
+      new RecordWithUnionAndTypeField(new ValidInt(255, t)).toByteBuffer
+    }
+  }
+
+  @State(Scope.Thread)
+  class Avro4sBytesSetup extends Setup {
+    val avro4sBytes = encode(RecordWithUnionAndTypeField(AttributeValue.Valid[Int](255, t)))
+  }
+
+  @State(Scope.Thread)
+  class AvroInputStreamSetup extends Setup {
+    private val outputStream = new ByteArrayOutputStream()
+    private val dataOutputStream = new AvroDataOutputStream[RecordWithUnionAndTypeField](outputStream, CodecFactory.nullCodec())
+    private val obj = RecordWithUnionAndTypeField(AttributeValue.Valid[Int](255, t))
+    dataOutputStream.write(0.to(255).map(i => RecordWithUnionAndTypeField(AttributeValue.Valid[Int](i, t))))
+    dataOutputStream.close()
+    private val bytes = outputStream.toByteArray
+
+    def avroDataStream =
+      new AvroDataInputStream[RecordWithUnionAndTypeField](new ByteArrayInputStream(bytes), None)
   }
 
   def encode[T: Encoder: SchemaFor](value: T): ByteBuffer = {
@@ -71,10 +90,14 @@ class Decoding extends CommonParams with BenchmarkHelpers {
   }
 
   @Benchmark
-  def avro4sHandrolled(setup: Setup, blackhole: Blackhole) =
+  def avro4sHandrolled(setup: Avro4sBytesSetup, blackhole: Blackhole) =
     blackhole.consume(decode(setup.avro4sBytes, setup.handrolledDecoder, setup.handrolledReader))
 
   @Benchmark
-  def avro4sGenerated(setup: Setup, blackhole: Blackhole) =
+  def avro4sGenerated(setup: Avro4sBytesSetup, blackhole: Blackhole) =
     blackhole.consume(decode(setup.avro4sBytes, setup.avro4sDecoder, setup.avro4sReader))
+  @Benchmark
+  def avroDataInputStream(setup: AvroInputStreamSetup, blackhole: Blackhole) = {
+    blackhole.consume(setup.avroDataStream.iterator.toList)
+  }
 }
--- a/benchmarks/src/main/scala/benchmarks/Decoding.scala
+++ b/benchmarks/src/main/scala/benchmarks/Decoding.scala
@@ -84,7 +84,7 @@ class Decoding extends CommonParams with BenchmarkHelpers {
 
 
   @Benchmark
-  def avroSpecificRecord(setup: Setup, blackhole: Blackhole) = {
+  def avroSpecificRecord(setup: AvroBytesSetup, blackhole: Blackhole) = {
     import benchmarks.record.generated._
     blackhole.consume(RecordWithUnionAndTypeField.fromByteBuffer(setup.avroBytes.duplicate))
   }


andreas-schroeder:
Hi @mberndt123,  thanks for your contribution! The simplifications are definitely welcome. 
Concerning the performance improvements, I share your position: I am not sure whether the allocation of GenericRecord actually makes a difference. How about we test it out? under /benchmarks you can find the infrastructure for creating JMH benchmarks that you can run with ./run_benchmarks.sh - note it might be necessary to run more than 10 warmup iterations and more than 10 passes to see the difference (or, if we're lucky, that isn't necessary because the difference is significant enough).

mberndt123:
Hey @andreas-schroeder,

I've added a benchmark for this. On my system, there is a small performance increase, it goes from 4596±115 ops/s to 4793±87 ops/s.

andreas-schroeder:
Hi @mberndt123, thanks for the benchmark extensions as they are don't compile, JMH complains about the type hierarchy of the setups. I've fixed them locally and did a pass with 1000 items with benchmarking
```scala
val it = setup.avroDataStream.iterator
    while(it.hasNext) {
      blackhole.consume(it.next)
    }
```
to avoid list allocation, and there's indeed a difference:
```
[info] Decoding.avroDataInputStreamOriginal1000  thrpt   30  3180,194 ± 37,733  ops/s
[info] Decoding.avroDataInputStreamReuse1000     thrpt   30  3479,936 ± 57,539  ops/s
```
even already with 10 items:
```
[info] Decoding.avroDataInputStreamOriginal10  thrpt   30  9118,079 ± 182,460  ops/s
[info] Decoding.avroDataInputStreamReuse10     thrpt   30  9272,318 ± 174,360  ops/s
```
So maybe we should reuse generic records in the other input stream types as well.

mberndt123:
Hey @andreas-schroeder ,

Oops, sorry for the compile error, it should work now.

mberndt123:
I've made the same change to AvroBinaryInputStream and AvroJsonInputStream.

sksamuel:
One thing to consider is with this change, a list generated from the stream will end up with the same single instance n times, all with the last records data.

mberndt123:
Hey @sksamuel,

The decoder is applied to the `GenericRecord` before a value is returned from `next()`, so you should still get distinct instances in that case, right?

Well, you can in theory write a `Decoder` that returns objects that keep a reference to the `GenericRecord` object, or a `Decoder[GenericRecord]` that just returns the record. But I don't think that's a common case.

sksamuel:
If you're passing in the original record, isn't the decoder just going to mutate that, otherwise where is the performance increase.

In which case, the reference you returned is going to be changed from under someone's feet.

andreas-schroeder:
Decoders for records are expected to read the GenericRecord and create a case class with values extracted from the record. The GenericRecord is just an intermediate container decoded from the binary or JSON data. The performance improvement, as far as I understand, comes from reusing this intermediate data container.

mberndt123:
Yes @andreas-schroeder, that's how it works.

andreas-schroeder:
Hi @mberndt123, I've made some small modifications here: https://github.com/mberndt123/avro4s/pull/1
* separating benchmarks
* running benchmark compilation on travis CI
* reorganize iterators to clarify record reuse

mberndt123:
Looks good to me.

mberndt123:
Is there anything left to do here to get this merged?

andreas-schroeder:
Looks good to me, @sksamuel what do you think?

sksamuel:
What if GenericRecord is what you want rather than a case class so no transformation takes place.

mberndt123:
In that case you'd have to make a copy of the `GenericRecord`.

sksamuel:
It could be a tricky bug for someone to track down if they were getting
records that were mutating under them.
Perhaps we could add a ClassTag to the InputStream, and if the type
parameter is GenericRecord we don't re-use, if it isn't we do ?

mberndt123:
Hey sksamuel,

I'm generally not a fan of this kind of special cases, and I'm even more sceptical about anything involving reflection. It destroys parametricity and it's also basically impossible to get right. You can handle the obvious case of `A =:= GenericRecord`, but what if `A` is some other class type that keeps a reference to the `GenericRecord`?

I only know of one way to solve this problem, which is to
1. make `Decoder` a `sealed trait` and make sure that all instances copy the data from the underlying `GenericRecord`. This would include a `Decoder` that simply copies the entire thing to a new instance.
1. add some combinators to combine these primitives into more complex ones. This would then be the only way to create new decoders.

If you structure things like that, it is not possible to create a decoder that retains a reference to the underlying mutable `GenericRecord`.

sksamuel:
ClassTags are added by the compiler at compile time.

But your suggestion of a sealed trait is good too.

mberndt123:
I'm closing this because @sksamuel has indicated he's not willing to merge it in its current form and I don't want to spend time changing the `Decoder` stuff in the way we discussed. A signature like `def decode(value: Any): T` is insane and not something I'm willing to work with.

sksamuel:
Well I was kinda waiting to see what you thought about my comment that ClassTag's are not runtime reflection, but compile time. Does that mitigate a lot of your objection ?

sksamuel:
What would be your ideal signature if you could rewrite the library from scratch ?

mberndt123:
I'm still don't think ClassTags are a good idea. One reason is what I pointed out earlier: it works for the obvious case of `A =:= GenericRecord`, but it doesn't work for e. g. `class Wrapper(x: GenericRecord)`. Another reason is that there are types for which there are no ClassTags, such as Newtypes. This is a way to define a new type that is opaque to the compiler while using the run-time representation of an existing type. The cats library contains several of these, specifically `cats.data.{NonEmptySet, NonEmptyChain, NonEmptyMap}`, and custom ones can be defined using e. g. [newts](https://julien-truffaut.github.io/newts/). Attempting to summon an implicit ClassTag for any of these will fail.

Regarding the signature of `read`: I like purely functional programming, so I think signatures should allow for an implementation that is defined sensibly for all possible inputs. This is impossible with `Any`, because `Any` can be, well, anything. The signature proposed by @andreas-schroeder in #498 is much better: `def readBytes(bytes: ByteBuffer, writerSchema: Schema): T`. For reasons already laid out in #498 it would be better to curry it, i. e. `def readBytes(writerSchema: Schema): ByteBuffer => T`. I'd also like a way to handle errors without exceptions, so `def readBytes(writerSchema: Schema): ByteBuffer => Either[DecodingError, T]`. Of course one would have to define a reasonable `DecodingError` type. I think this would be a useful definition for future versions of avro4s.

If I really wanted to start from scratch, I'd also like to allow side effects in a controlled manner while not constraining to any effect type in particular.
```
trait Decoder[F[_, _], A] {
  def decode(writerSchema: Schema): ByteBuffer => F[DecodingError, A]
}
```
Possible `F[_, _]` types would be `Either`, `ZIO[Any, *, *]` (using zio) or `EitherT[IO, *, *]` (using cats-effect).

sksamuel:
So say we have a `StringDecoder` which can decode a value from a field in an Avro record into a String.

The implementation supports `GenericFixed`, `String`, `Utf8`, `Array[Byte]` and so on.

The signature using ByteBuffer isn't compatible here. That's why the reason for Any.

If we had something like:

```scala
trait Decoder[A, B] {
   def decode(a: A): B
}
```

Then we could have something like:

`object GenericFixedStringDecoder extends Decoder[GenericFixed, String] { 
  def decode(a: GenericFixed): String
}`

This would be nice but how would we summon that at compile time as the input records might be anything. In fact the same avro document might contain UTF8s and java Strings and GenericFixed's interspersed.

On the other hand, I can see the argument for adding effect types to the IO interfaces as well as functional error handling. Ideas welcome.

mberndt123:
If I understand correctly, there's only a finite set of types that need to be considered. How about a wrapper ADT?

```scala
sealed trait AvroValue
case class AvroString(s: String) extends AvroValue
case class AvroFixed(bytes: Array[Byte] extends AvroValue
class AvroRecord(underlying: GenericRecord) extends AvroValue {
  def get(name: String): Option[AvroValue] = Option(underlying.get(name)).map(AvroValue.unsafeFromAny)
  def get(i: Int): Option[AvroValue] = Option(underlying.get(i)).map(AvroValue.unsafeFromAny)
}
// etc.

object AvroValue {
  private[avro4s] unsafeFromAny(a: Any): AvroValue = a match {
    case r: GenericRecord => new AvroRecord(r)
    case s: String => AvroString(s)
    case f: GenericFixed => AvroFixed(f.bytes)
    // etc.
  }
}
```

Then the Decoder interface could take an `AvroValue` instead of `Any`.

The downside is that this would require more allocations at runtime… But it might be possible to avoid this using some type-level trickery.

sksamuel:
I've done something similar to that before in other frameworks I've written.

This would seem to cover it.

```scala
object AvroValue {

  import scala.collection.JavaConverters._

  case class AvroString(s: String) extends AvroValue
  case class AvroUtf8(utf8: Utf8) extends AvroValue
  case class AvroByteArray(bytes: Array[Byte]) extends AvroValue
  case class AvroGenericFixed(fixed: GenericFixed) extends AvroValue
  case class AvroCharSequence(charseq: CharSequence) extends AvroValue
  case class AvroBoolean(boolean: Boolean) extends AvroValue
  case class AvroShort(short: Short) extends AvroValue
  case class AvroByte(byte: Byte) extends AvroValue
  case class AvroDouble(double: Double) extends AvroValue
  case class AvroFloat(float: Float) extends AvroValue
  case class AvroInt(int: Int) extends AvroValue
  case class AvroLong(long: Long) extends AvroValue
  case class AvroEnumSymbol(symbol: GenericEnumSymbol[_]) extends AvroValue
  case class AvroList(list: List[AvroValue]) extends AvroValue

  class AvroRecord(underlying: GenericRecord) extends AvroValue {
    def get(name: String): Option[AvroValue] = Option(underlying.get(name)).map(AvroValue.unsafeFromAny)
    def get(i: Int): Option[AvroValue] = Option(underlying.get(i)).map(AvroValue.unsafeFromAny)
  }

  private[avro4s] def unsafeFromAny(a: Any): AvroValue = a match {
    case r: GenericRecord => new AvroRecord(r)
    case s: String => AvroString(s)
    case f: Utf8 => AvroUtf8(f)
    case g: GenericFixed => AvroGenericFixed(g)
    case b: ByteBuffer => AvroByteArray(b.array())
    case a: Array[Byte] => AvroByteArray(a)
    case c: java.lang.Iterable[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
    case c: java.util.Collection[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
    case a: Array[_] => AvroList(a.toList.map(AvroValue.unsafeFromAny))
    case f: Float => AvroFloat(f)
    case d: Double => AvroDouble(d)
    case b: Byte => AvroByte(b)
    case s: Short => AvroShort(s)
    case i: Int => AvroInt(i)
    case l: Long => AvroLong(l)
    case e: GenericEnumSymbol[_] => AvroEnumSymbol(e)
    case _ => throw new Avro4sUnsupportedTypeException(s"$a is not a supported type when decoding")
  }
}
```

I don't know if we'd want to include AvroNull in there as a type, or handle it elsewhere.

andreas-schroeder:
Before going further with this, I would suggest to check whether we could use `AnyVal`'s to avoid object allocations, prototype how the record encoding / decoding would need to be adapted, and measure the performance overhead of the change.

sksamuel:
Oops well I went ahead and made the change (was quick enough to do).
https://github.com/sksamuel/avro4s/pull/535

But yes, we shouldn't consider merging until we know the performance cost.

mberndt123:
The code in #535 sure looks a lot better than what was there before! Let's hope it doesn't impact performance too much…

But regarding this Pull Request, I'm not sure if it's enough.

When you want a compositional API like I have suggested earlier, you need some operations to build new Decoders from existing ones, e. g.
```scala
trait Decoder[A] {
  def zip[B](d: Decoder[B]): Decoder[(A, B)] = new Decoder[(A, B)] { … }
  …
}
```
The decoding part is fairly straight-forward, but what do you do about `schemaFor`? It's not obvious to me what the composite schema should look like, and I don't understand why it's needed in the first place. 

My understanding was that Avro uses the reader/writer schema to figure out how to build a `GenericRecord` object from the binary data, and the `Decoder[A]`'s job is then to transform the `GenericRecord` to an object of type `A`? Is that correct? Because if it is, then I don't see why a `Decoder` would need to know about the schema at all.

sksamuel:
The reason for the schema is because just knowing the types isn't enough to do the decoding.
For example, big decimals have a scale and precision that isn't obvious from the byte buffer. For that you need the schema.
Similarly, you need the schema a default value if you have a union. There's plenty of such examples.
This is why avro4s has a much more complicated codebase than something like a json parser, because of the schema step that goes with it.

mberndt123:
Thanks for explaining, that is interesting. You're apparently referring to [Logical Types](https://avro.apache.org/docs/1.10.0/spec.html#Logical+Types). The documentation states that 

> Language implementations may choose to represent logical types with an appropriate native type, although this is not required.

That made me think: if they explicitly point this out in their documentation, then surely they have implemented it in their own Java implementation, right? 

To me it looks like we can get a [`GenericData`](https://avro.apache.org/docs/current/api/java/org/apache/avro/generic/GenericData.html) instance and add the relevant conversions using [`addLogicalTypeConversion`](https://avro.apache.org/docs/current/api/java/org/apache/avro/generic/GenericData.html#addLogicalTypeConversion-org.apache.avro.Conversion-). Then you get a `DatumReader` using [`createDatumReader`](https://avro.apache.org/docs/current/api/java/org/apache/avro/generic/GenericData.html#createDatumReader-org.apache.avro.Schema-org.apache.avro.Schema-) and use that to create the [`DataFileStream`](http://avro.apache.org/docs/current/api/java/org/apache/avro/file/DataFileStream.html#DataFileStream-java.io.InputStream-org.apache.avro.io.DatumReader-). I would presume that things like `BigDecimal` would then be decoded according to the Schema automatically and `GenericRecord.get` would simply return a `BigDecimal` instance. Is that how it works, or am I completely off-track here?

mberndt123:
Hey @sksamuel,

I've thrown together some code to test this:

```scala
import com.sksamuel.avro4s
import com.sksamuel.avro4s._
import org.apache.avro.file.CodecFactory
import java.io._
import org.apache.avro.generic._
import org.apache.avro.Conversions
import org.apache.avro.file.DataFileStream

object Main extends App {
  case class Foo(bd: BigDecimal)
  val schema = AvroSchema[Foo]

  val os = new AvroDataOutputStream[Foo](new FileOutputStream("/tmp/foo"), CodecFactory.nullCodec())
  try {
    os.write(Foo(BigDecimal(42)))
  } finally os.close()
  val is = new AvroDataInputStream[Foo](new FileInputStream("/tmp/foo"), None)

  val genericData = new GenericData
  genericData.addLogicalTypeConversion(new Conversions.DecimalConversion)
  val datumReader = genericData.createDatumReader(SchemaFor[Foo].schema, SchemaFor[Foo].schema)
  val dataFileStream: DataFileStream[_] = new DataFileStream(new FileInputStream("/tmp/foo"), datumReader)
  val record = try {
    dataFileStream.iterator().next()
  } finally dataFileStream.close
  assert(record.asInstanceOf[GenericRecord].get("bd").getClass.getSimpleName == "BigDecimal")
}
```

The assertion does indeed succeed, so apparently my understanding of how this works is correct. What do you think, is it possible to just let Avro take care of the schema entirely? Because at least for the `BigDecimal` case it seems to work…

sksamuel:
Well it certainly makes sense. Let me run some other scenarios to double
check you're right, but assuming you are, how would you envisage things
changing?
It's quite a major change and Andreas has done a lot of legwork getting
performance up so we can't lose that.

mberndt123:
Well, the idea would be to decouple `Decoder` entirely from the Schema, meaning that `SchemaAware` would just go away, because Avro itself would take care of everything having to do with the Schema. But I don't know if that will actually work, and you're probably in a better position to know, having written the library :smile:. All I can say is that that approach seems to work for the one case you've mentioned where the schema is required, i. e. BigDecimal.

Regarding the `AvroValue` in #535, it would need to be extended to also comprise all the Logical types in the Avro documentation – BigDecimal, UUID etc..

Regarding performance, I think the essential “special sauce” here is how avro4s avoids the lazy unrolling of the type class definition for cyclic types, right? I don't see how that would be affected by such a change.

sksamuel:
The special unrolling is what Andreas has brought in as well as a ton of
massive perf enhancements.

I'll think about schemas in Decoder. There's also Encoder. Have you
considered that. How do you encode a big decimal without the schema.

andreas-schroeder:
The other relevant change I did was to make sure that the schema-specific encoding/decoding cusomization was performed only once and not on every encoding/decoding operation. The challenge with recursive types was to ensure this change could be carried through even with recursive types.

However looking back at the original goal of creating a composable api, I have a hard time seeing the issue with SchemaFor. If there is a composition schema for type classes that works for decoder, what makes it break with SchemaFor? On the other hand, we will have to offer the same composable api for encoders as well, and here we have to address the issue of SchemaFor compostability.

mberndt123:
Hey @andreas-schroeder,

I think the Schema stuff does present a problem for a composable `Decoder` API. Take for instance this example from the Circe documentation:
```
implicit val decodeFoo: Decoder[Thing] = new Decoder[Thing] {
  final def apply(c: HCursor): Decoder.Result[Thing] =
    for {
      foo <- c.downField("foo").as[String]
      bar <- c.downField("bar").as[Int]
    } yield {
      new Thing(foo, bar)
    }
}
```
This is clearly a decoder for a record type called `Thing`, but with the current `Decoder` definition it's not possible to support such an API, because the resulting `Decoder` would need to contain a schema with this named record type. You can't create such a schema from the above code, because that would require the name `"Thing"` as a string from somewhere, and that string doesn't appear in the above code. Sure, it's probably possible to make that work with some sort of black magic (macros, introspection), but I think it would be better to avoid the problem in the first place. Why have the Schema when the `DatumReader` can sort out things like `BigDecimal` before the `Decoder` even gets to see the data?

Regarding the `Encoder`, it would work exactly like `Decoder`, except the other way around. You create the usual data structure using `GenericRecord`, `GenericData.Array` etc., and if you need to encode a `BigDecimal`, you stick it right in there. The Avro `DatumWriter` will find it and encode it according to the Schema that you supplied when you created the `DatumWriter`.

Of course you still need the `SchemaFor` to do any reading and writing, but I would just pass it as a separate implicit paramter when creating `AvroDataInputStream`, `AvroDataOutputStream` etc. 

Of course I might be missing something here, so feel free to comment and explain to me how nothing I wrote makes any sense :wink:

sksamuel:
You make it all sound so simple so I feel like it must be more complicated
:)

sksamuel:
So just been playing around with the low level code again as a refresher from 2015 when I first laid down the foundations for these basic typeclasses.

Let's say I have a class `case class Foo(a: String)` and I want to create a `GenericRecord`. In this record I want my string to be represented by a `GenericFixed`.

At the moment I can use the `RecordEncoder` typeclass passing in a schema (although you wouldn't invoke it directly, you'd use `ToRecord`), and that schema gets passed in turn to the `StringEncoder` which will then use that schema to decide how to marshall the string.

```scala
    val encoder: String => AnyRef = schema.getType match {
      case Schema.Type.STRING => new Utf8(_)
      case Schema.Type.FIXED  => encodeFixed
      case Schema.Type.BYTES =>
        str =>
          ByteBuffer.wrap(str.getBytes)
      case _ => throw new Avro4sConfigurationException(s"Unsupported type for string schema: $schema")
    }
```

If the `Encoder` trait was changed to be schemaless:

```scala
trait Encoder[T] extends Serializable {
  def encode(value: T): AvroValue
}
```

How would we do this? I still need a schema to create the `GenericRecord` instance (since it's part of the constructor).
I can use the DatumWriter to write out to a stream but it doesn't help me get a `GenericRecord` instance to start with ?

andreas-schroeder:
For records, wouldn't that be what we built in #494 ? For Avro you sure have to provide much more additional info than for Json (namespace, name, doc, field name, field doc, ...) but then that's it. For defining tuple typeclasses via zip, we could just delegate to the code that already exists for typeclass derivation of tuples, couldn't we?

sksamuel:
If I have a type, `Foo` and I want to automagically create a `GenericRecord` (the goal of Avro4s), then I want to do something like:

`val record = ToRecord[Foo](Foo(....))`

If Foo has a field of type String that I want to marshall as a GenericFixed rather than a Utf8, how do I do that without specifying the schema somewhere and passing that schema down to a String Encoder.

mberndt123:
Hey @sksamuel,

regarding the different possible Encoders for String (Fixed, Bytes, String), I think this problem shouldn't be hard to solve. After all, the `SchemaFor` instance and the `Encoder` instance are generated from the same underlying source of information: the case class declaration and, more specifically, the annotation therein:

```
case class Foo(@AvroFixed(7) mystring: String)
```

It's possible to customize Magnolia derivations based on annotations, and it shouldn't be a problem to generate an `Encoder` for `Foo` that encodes `myString` as a `GenericFixed`. It should even be a bit faster, because the decision on how to encode the String will be done at compile time, not at run-time.

sksamuel:
The encoder instance doesn't just use the schema from the case class. The schema can come from anywhere. Typically it would be from the case class for people doing entirely automatic derivation, but it can be passed in at runtime - a schema generated elsewhere. We would need to allow a way to do that I think before we can ditch schemas in the Encoders.

mberndt123:
Well in that case the Encoder would need a signature like
```trait Encoder[T] {
  def encode(schema: Schema): T => AvroValue
}
```
But I think we've reached a point in the discussion where it starts to look more and more like a rewrite… I don't think it makes sense to pursue this further for the 4.0.0 release.

sksamuel:
That's a shame. I had come around to agreeing with you on the merits of the
new signatures.

Perhaps we could visit this for the dotty version.

sksamuel:
And we could still adopt the AvroValue ADT.

mberndt123:
Sure, the AvroValue thing might still make sense if the performance issues can be resolved.

sksamuel:
It's about the same performance now.

mberndt123:
Oh nice. Glad to hear that that part worked out!

PR #516, opened 2020-07-09
Check whether class is value class

### LINKED ISSUE #509: AvroDoc annotation will result in multiple doc fields
Given this message structure:
```
@AvroDoc("This is record 1")
case class Record1(field: String)
case class Message(record1: Record1)
```

Avro4s will generate the following schema:
```
{
  "type": "record",
  "name": "Message",
  "namespace": "Avro4s",
  "fields": [
    {
      "name": "record1",
      "type": {
        "type": "record",
        "name": "Record1",
        "doc": "This is record 1",
        "fields": [
          {
            "name": "field",
            "type": "string"
          }
        ]
      },
      "doc": "This is record 1"
    }
  ]
}
```

I would like to have it generate a schema where the doc field for Record1 is only in the object where Record1 is defined like this:
```
{
  "type": "record",
  "name": "Message",
  "namespace": "Avro4s",
  "fields": [
    {
      "name": "record1",
      "type": {
        "type": "record",
        "name": "Record1",
        "doc": "This is record 1",
        "fields": [
          {
            "name": "field",
            "type": "string"
          }
        ]
      }
    }
  ]
}
```

Any ideas?

Some background:
We use the io.confluent.kafka.serializers.KafkaAvroSerializer for serializing the message which requires the schema to exactly match the one registered in the Schema Registry (including the doc fields). Since we have multiple producers (over multiple teams) for the same topic we defined a schema centrally and manually upload it to the Schema Registry. This schema is defined like my second example. In this specific producer we use Avrohugger to create the case classes, then manually add the AvroDocs annotations and use Avro4s to transform the case class in a GenericRecord to be used with the KafkaAvroSerializer.

Reproduction:

```
import com.sksamuel.avro4s.{AvroDoc, ToRecord}

object Avro4s extends App {
  @AvroDoc("This is record 1")
  case class Record1(field: String)
  case class Message(record1: Record1)

  val derivedSchema = ToRecord[Message].to(Message(Record1("value"))).getSchema

  println(derivedSchema)
}
```

andreas-schroeder:
In order to avoid duplicating the doc field, check whether the current class is a value class before extracting the annotation from it.

Fixes #509

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/Records.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/Records.scala
@@ -142,9 +142,6 @@ object Records {
       if (annotations.transient) None
       else {
         val doc = valueTypeDoc(ctx, param)
-        if(doc.isDefined) {
-          println("context with doc: " + ctx.typeName.full)
-        }
         val schema = param.typeclass.resolveSchemaFor(nextEnv, NoUpdate).schema
         Some(buildSchemaField(param, schema, annotations, record.getNamespace, fieldMapper, doc))
       }
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/Records.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/Records.scala
@@ -142,6 +142,9 @@ object Records {
       if (annotations.transient) None
       else {
         val doc = valueTypeDoc(ctx, param)
+        if(doc.isDefined) {
+          println("context with doc: " + ctx.typeName.full)
+        }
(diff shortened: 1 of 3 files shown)


avro4s-core/src/main/scala/com/sksamuel/avro4s/Records.scala
```diff
@@ -142,6 +142,9 @@ object Records {
       if (annotations.transient) None
       else {
         val doc = valueTypeDoc(ctx, param)
+        if(doc.isDefined) {
+          println("context with doc: " + ctx.typeName.full)
```
sksamuel, line 146: debugging left in

Issue #495, opened 2020-06-02
Significant performance regression between version 3.0.9 and 2.03


KGoodale13:
Hello, we recently upgraded from avros4s version 2.0.3 to 3.0.9 and have noticed some massive performance regressions. 

Some basic benchmarks show a hit of about 3X on serialization: 
```
Avro Serialization (3.0.9):
Parameters(numberOfEvents -> 500): 105.949524 ms
Parameters(numberOfEvents -> 2000): 426.090214 ms
Parameters(numberOfEvents -> 3500): 745.042044 ms
Parameters(numberOfEvents -> 5000): 1061.600599 ms
Parameters(numberOfEvents -> 6500): 1379.572099 ms
```

```
Avro Serialization (2.0.3):
Parameters(numberOfEvents -> 500): 32.817919 ms
Parameters(numberOfEvents -> 2000): 134.431979 ms
Parameters(numberOfEvents -> 3500): 238.911994 ms
Parameters(numberOfEvents -> 5000): 333.905703 ms
Parameters(numberOfEvents -> 6500): 435.456092 ms
```

Independent of those benchmarks we have observed an even greater hit to cpu and serialization times when using avro4s 3.0.9 in production.
![Screen Shot 2020-06-02 at 2 35 50 PM](https://user-images.githubusercontent.com/13970273/83557555-d9629400-a4df-11ea-872e-d087a0cc99a8.png)
In the screenshot above we have a particular service running avro4s version 3.0.9 on the left side. After reverting to avro4s version 2.0.3 with no other changes you can see the serialization time drop significantly on the right side.

Additional info: 
- This only appears to effect serialization times. Deserialization times remain unaffected based on our observations. 
- We are serializing to a binary avro output stream.

andreas-schroeder:
Thank you for the detailed report. Yes, there is a significant performance issue and we have been working on fixing it in version 4.0.0 for a couple weeks now. Maybe the new version will bring you the performance back that you need, maybe even better. I'd love to have your numbers with version 4.0.0 once it is there.

sksamuel:
Can you try 4.0.0.RC1 now it has been released @KGoodale13

KGoodale13:
It appears it may have gotten worse. I did five separate runs to confirm. Here are the results for both serialization and deserialization for versions 4.0.0-RC1 and 2.0.3:
```
4.0.0-RC1 Avro Serialization
[...]

PR #519, opened 2020-07-17
Use ad-hoc instances so that expensive resolving isn't performed on stream creation time

andreas-schroeder:
This is the easiest way to avoid the performance hit discussed [here](https://github.com/sksamuel/avro4s/issues/495#issuecomment-659492364) 
we can fall back to use the ad-hoc instance of the resolvable encoder / decoder, which is a lazy val and has some slight performance hit, but by this it becomes impossible to accidentally perform resolution multiple times.

I haven't had time to create benchmarks yet to check how much of a win it would be to explicitly resolve at the right place, I will investigate in the next couple days. I expect the win to be very small, if noticeable at all, to be honest.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroBinaryInputStream.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroBinaryInputStream.scala
@@ -20,9 +20,7 @@ class AvroBinaryInputStream[T](in: InputStream,
                                writerSchema: Schema)
                               (implicit decoder: Decoder[T]) extends AvroInputStream[T] {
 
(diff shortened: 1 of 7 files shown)


sksamuel:
Why does calling .resolveDecoder once in the constructor make it happen multiple times ?

andreas-schroeder:
Sure, it happens once per instance, but the stream instances are created once per build() call. So in total resolveX are called multiple times, worst case once per record to decode / encode. That's what I meant by "multiple times".

sksamuel:
I see, because in the original code, a new output stream is created each time.
Could always do the resolve call in the Stream Builder rather than the stream itself, but probably would be small gain.

Issue #502, opened 2020-06-17
serialization fails with named values in enums


mdespriee:
Using scala 2.13.1, avro4s-core 3.1.1

When naming values in enums, the serializer fails with:
```
java.lang.NullPointerException: null of com.sencrop.serialization.NamedEnum in field m of com.sencrop.serialization.Foo
	at org.apache.avro.generic.GenericDatumWriter.npe(GenericDatumWriter.java:183)
	at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:177)
	at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:82)
	at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:72)
	at com.sksamuel.avro4s.DefaultAvroOutputStream.write(DefaultAvroOutputStream.scala:22)
 ...
Caused by: java.lang.NullPointerException
	at org.apache.avro.Schema$EnumSchema.getEnumOrdinal(Schema.java:1025)
	at org.apache.avro.generic.GenericDatumWriter.writeEnum(GenericDatumWriter.java:219)
	at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:133)
	at org.apache.avro.generic.GenericDatumWriter.write(GenericDatumWriter.java:82)
	at org.apache.avro.generic.GenericDatumWriter.writeField(GenericDatumWriter.java:206)
	at org.apache.avro.generic.GenericDatumWriter.writeRecord(GenericDatumWriter.java:195)
	at org.apache.avro.generic.GenericDatumWriter.writeWithoutConversion(GenericDatumWriter.java:130)
	... 54 more
```

Reproducer:
```

import com.sksamuel.avro4s.{AvroOutputStream, SchemaFor}
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.should.Matchers

object ClassicEnum extends Enumeration {
  val T = Value
}
case class Bar(m: ClassicEnum.Value)

object NamedEnum extends Enumeration {
  val T = Value("Temperature")
}
case class Foo(m: NamedEnum.Value)

class BugDemo extends AnyFlatSpec with Matchers {

  "serialize" should "work with classic enum" in {
    println(SchemaFor[Bar].schema.toString(true))
    val byteStr = new java.io.ByteArrayOutputStream
[...]

PR #534, opened 2020-08-10
Added support for scala named enums in schemas #502

sksamuel:
This could be tidied up further. Mostly I would prefer to use the mirror on the type tag to obtain the module instance but I couldn't figure that out.


sksamuel:
These tests pass for me locally. I will investigate more tomorrow.

PR #535, opened 2020-08-10
Introduced AvroValue ADT and updated Decoder interface

sksamuel:
Added `AvroValue` ADT and changed `Decoder` trait to define `def decode(value: AvroValue): T`

Some points:

* I decided to do the normalization of values at the time we build the ADT. For example for strings, possibly "stringy" types are CharSequence, String and UTF8. Rather than define 3 types in the ADT, I just normalize to a single type, AvroString. Another example is instead of two types for byte arrays and ByteBuffer, I normalize both to AvroByteArray.
* The decoders are simplified because they no longer need to match on multiple possible "java" types.
* For records, I defined AvroRecord which wraps GenericRecord. I don't know if we want to wrap IndexedRecord instead, or even have multiple types in the ADT.
* This change only covers Decoder and not Encoders.
* All tests are passing but I have **not** measured performance.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
@@ -1,0 +1,63 @@
+package com.sksamuel.avro4s
+
+import java.nio.ByteBuffer
+
+import org.apache.avro.generic.{GenericEnumSymbol, GenericFixed, GenericRecord}
+import org.apache.avro.util.Utf8
+
+sealed trait AvroValue
+
+object AvroValue {
+
+  import scala.collection.JavaConverters._
+
+  case class AvroString(s: String) extends AvroValue
+  case class AvroByteArray(bytes: Array[Byte]) extends AvroValue
+  case class AvroGenericFixed(fixed: GenericFixed) extends AvroValue
+  case class AvroBoolean(boolean: Boolean) extends AvroValue
+  case class AvroShort(short: Short) extends AvroValue
+  case class AvroByte(byte: Byte) extends AvroValue
+  case class AvroDouble(double: Double) extends AvroValue
+  case class AvroFloat(float: Float) extends AvroValue
+  case class AvroInt(int: Int) extends AvroValue
+  case class AvroLong(long: Long) extends AvroValue
+  case class AvroEnumSymbol(symbol: GenericEnumSymbol[_]) extends AvroValue
+  case class AvroList(list: List[AvroValue]) extends AvroValue
+  case class AvroMap(map: Map[String, AvroValue]) extends AvroValue
+  case object AvroNull extends AvroValue
+
+  case class AvroRecord(record: GenericRecord) extends AvroValue {
+    def get(name: String): Option[AvroValue] = Option(record.get(name)).map(AvroValue.unsafeFromAny)
+    def get(i: Int): Option[AvroValue] = Option(record.get(i)).map(AvroValue.unsafeFromAny)
+  }
+
+  private[avro4s] def unsafeFromAny(a: Any): AvroValue = a match {
+    case r: GenericRecord => AvroRecord(r)
+    case s: String => AvroString(s)
+    case u: Utf8 => AvroString(u.toString)
+    case c: CharSequence => AvroString(c.toString)
+    case b: Boolean => AvroBoolean(b)
+    case g: GenericFixed => AvroGenericFixed(g)
+    case b: ByteBuffer => AvroByteArray(b.array())
+    case a: Array[Byte] => AvroByteArray(a)
+    case a: Array[_] => AvroList(a.toList.map(AvroValue.unsafeFromAny))
+    case m: java.util.Map[_, _] => AvroMap(m.asScala.map { case (key, value) => key.toString -> unsafeFromAny(value) }.toMap)
+    case c: java.lang.Iterable[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
+    case c: java.util.Collection[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
+    case f: Float => AvroFloat(f)
+    case f: java.lang.Float => AvroFloat(f)
+    case d: Double => AvroDouble(d)
+    case d: java.lang.Double => AvroDouble(d)
+    case b: Byte => AvroByte(b)
+    case b: java.lang.Byte => AvroByte(b)
+    case s: Short => AvroShort(s)
+    case s: java.lang.Short => AvroShort(s)
+    case i: Int => AvroInt(i)
+    case i: java.lang.Integer => AvroInt(i)
+    case l: Long => AvroLong(l)
+    case l: java.lang.Long => AvroLong(l)
+    case e: GenericEnumSymbol[_] => AvroEnumSymbol(e)
+    case null => AvroNull
+    case _ => throw new Avro4sUnsupportedValueException(s"$a is not a supported type when decoding. All types must be wrapped in AvroValue.")
+  }
+}
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
@@ -3,61 +3,50 @@ package com.sksamuel.avro4s
 import java.nio.ByteBuffer
 
 import org.apache.avro.generic.{GenericEnumSymbol, GenericFixed, GenericRecord}
-import org.apache.avro.util.Utf8
 
-sealed trait AvroValue
+import scala.collection.JavaConverters._
+
+sealed trait AvroValue 
 
 object AvroValue {
 
-  import scala.collection.JavaConverters._
-
-  case class AvroString(s: String) extends AvroValue
-  case class AvroByteArray(bytes: Array[Byte]) extends AvroValue
-  case class AvroGenericFixed(fixed: GenericFixed) extends AvroValue
-  case class AvroBoolean(boolean: Boolean) extends AvroValue
-  case class AvroShort(short: Short) extends AvroValue
-  case class AvroByte(byte: Byte) extends AvroValue
-  case class AvroDouble(double: Double) extends AvroValue
-  case class AvroFloat(float: Float) extends AvroValue
-  case class AvroInt(int: Int) extends AvroValue
-  case class AvroLong(long: Long) extends AvroValue
-  case class AvroEnumSymbol(symbol: GenericEnumSymbol[_]) extends AvroValue
-  case class AvroList(list: List[AvroValue]) extends AvroValue
-  case class AvroMap(map: Map[String, AvroValue]) extends AvroValue
-  case object AvroNull extends AvroValue
+  case class AvroString(str: String) extends AvroValue 
+  case class AvroByteArray(bytes: Array[Byte]) extends AvroValue 
+  case class AvroGenericFixed(fixed: GenericFixed) extends AvroValue 
+  case class AvroBoolean(boolean: Boolean) extends AvroValue 
+  case class AvroShort(short: Short) extends AvroValue 
+  case class AvroByte(byte: Byte) extends AvroValue 
+  case class AvroDouble(double: Double) extends AvroValue 
+  case class AvroFloat(float: Float) extends AvroValue 
+  case class AvroInt(int: Int) extends AvroValue 
+  case class AvroLong(long: Long) extends AvroValue 
+  case class AvroEnumSymbol(symbol: GenericEnumSymbol[_]) extends AvroValue 
+  case class AvroList(list: List[AvroValue]) extends AvroValue 
+  case class AvroMap(map: Map[String, AvroValue]) extends AvroValue 
+  case object AvroNull extends AvroValue 
 
   case class AvroRecord(record: GenericRecord) extends AvroValue {
     def get(name: String): Option[AvroValue] = Option(record.get(name)).map(AvroValue.unsafeFromAny)
     def get(i: Int): Option[AvroValue] = Option(record.get(i)).map(AvroValue.unsafeFromAny)
   }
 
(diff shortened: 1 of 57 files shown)


sksamuel:
I cannot add you as a reviewer @mberndt123 so am tagging you here.

mberndt123:
Wow @sksamuel, I wasn't expecting such swift action just because I moaned a bit about that signature! At a first glance it looks great. I'll take a closer look as soon as I have time.

sksamuel:
Sure that would be good. If we do decide to do this, we should do it
quickly to get it into 4.0.0.
After 4.0.0 is released, we can't make any breaking changes.
Do need to confirm performance and see if I can use value types.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
```diff
@@ -0,0 +1,63 @@
...
+    case u: Utf8 => AvroString(u.toString)
+    case c: CharSequence => AvroString(c.toString)
+    case b: Boolean => AvroBoolean(b)
+    case g: GenericFixed => AvroGenericFixed(g)
+    case b: ByteBuffer => AvroByteArray(b.array())
+    case a: Array[Byte] => AvroByteArray(a)
+    case a: Array[_] => AvroList(a.toList.map(AvroValue.unsafeFromAny))
+    case m: java.util.Map[_, _] => AvroMap(m.asScala.map { case (key, value) => key.toString -> unsafeFromAny(value) }.toMap)
+    case c: java.lang.Iterable[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
+    case c: java.util.Collection[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
```
mberndt123, line 46: `Collection` is a subtype of `Iterable`, so I think this is case is unreachable

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
```diff
@@ -0,0 +1,63 @@
...
+  case class AvroGenericFixed(fixed: GenericFixed) extends AvroValue
+  case class AvroBoolean(boolean: Boolean) extends AvroValue
+  case class AvroShort(short: Short) extends AvroValue
+  case class AvroByte(byte: Byte) extends AvroValue
+  case class AvroDouble(double: Double) extends AvroValue
+  case class AvroFloat(float: Float) extends AvroValue
+  case class AvroInt(int: Int) extends AvroValue
+  case class AvroLong(long: Long) extends AvroValue
+  case class AvroEnumSymbol(symbol: GenericEnumSymbol[_]) extends AvroValue
+  case class AvroList(list: List[AvroValue]) extends AvroValue
```
mberndt123, line 25: Is it necessary to allocate a list here? We could avoid many allocations by making this an `Iterable[AvroValue]` and merely wrapping the underlying collection.
This would also be analogous to `AvroRecord` where also also just wrap the underlying `GenericRecord`.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
```diff
@@ -0,0 +1,63 @@
...
+  case class AvroMap(map: Map[String, AvroValue]) extends AvroValue
+  case object AvroNull extends AvroValue
+
+  case class AvroRecord(record: GenericRecord) extends AvroValue {
```
mberndt123, line 29: If you declare it as a `case class`, it is not possible to make `record` private. But I think that would be desirable to hide the Java nastiness away from where it's not needed.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
```diff
@@ -0,0 +1,63 @@
...
+  case class AvroBoolean(boolean: Boolean) extends AvroValue
+  case class AvroShort(short: Short) extends AvroValue
+  case class AvroByte(byte: Byte) extends AvroValue
+  case class AvroDouble(double: Double) extends AvroValue
+  case class AvroFloat(float: Float) extends AvroValue
+  case class AvroInt(int: Int) extends AvroValue
+  case class AvroLong(long: Long) extends AvroValue
+  case class AvroEnumSymbol(symbol: GenericEnumSymbol[_]) extends AvroValue
+  case class AvroList(list: List[AvroValue]) extends AvroValue
+  case class AvroMap(map: Map[String, AvroValue]) extends AvroValue
```
mberndt123, line 26: Like for lists, I think it would be better to avoid intermediate collections here.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
```diff
@@ -0,0 +1,63 @@
...
+    case b: Boolean => AvroBoolean(b)
+    case g: GenericFixed => AvroGenericFixed(g)
+    case b: ByteBuffer => AvroByteArray(b.array())
+    case a: Array[Byte] => AvroByteArray(a)
+    case a: Array[_] => AvroList(a.toList.map(AvroValue.unsafeFromAny))
+    case m: java.util.Map[_, _] => AvroMap(m.asScala.map { case (key, value) => key.toString -> unsafeFromAny(value) }.toMap)
+    case c: java.lang.Iterable[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
+    case c: java.util.Collection[_] => AvroList(c.asScala.map(AvroValue.unsafeFromAny).toList)
+    case f: Float => AvroFloat(f)
+    case f: java.lang.Float => AvroFloat(f)
```
mberndt123, line 48: I think the `java.lang` variants aren't really necessary here. `a` is `Any` anyway, so it's always going to be boxed.

mberndt123:
OK, I think this is a much better foundation to build stuff on than what we had originally.

So where do we go from here? The original reason we even started working on this is that I was thinking about a more compositional API in #499. But I think this would require quite a few additional changes, and I wonder if you want to make so many changes at this point. The current release is 4.0.0-RC2, so you were apparently expecting to cut a release fairly soon, and I don't want to disrupt that too much.

sksamuel:
Under normal circumstances I'd say it's too late in the cycle to make substantial changes, but if we're changing the signature of the Decoder and Encoder (and obviously we may still not), then 4.0 is the only time to do it.

I like the AvroValue and it cleaned up a lot of matching code inside the decoders.

There will probably never be a 5.0 for Scala, as the next major work would be a Dotty release.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
sksamuel: That could work and would save iterating things twice + allocations.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
sksamuel: It was nicer to pattern match which is why I did it. Might there be a use case to have access to the actual record?

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
sksamuel: And just wrap the raw Java map ?

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
sksamuel: I definitely had an issue in the past, which is why all the decoders matched on both java.lang and scala types. Perhaps that has gone away, but leaving it in won't hurt.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
mberndt123: Yes

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
mberndt123: Well, you might have already written some code to parse a data structure from a `GenericRecord`, and if you want to reuse that, you're going to need access to the `GenericRecord`.

My original thought was that we want to keep `GenericRecord` out of the hands of users, because they might stow it away and read it later, and then they would get weird results if the `GenericRecord` was overwritten (#499). But thinking about it, they can do just the same thing with the `AvroRecord` itself, which would lead to the same result (unless `AvroRecord` creates a copy, which would be bad for performance reasons). So actually I think a `case class` is fine here.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
mberndt123: yes, that was my thinking, too

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroValue.scala
andreas-schroeder: The Java types are what you get back as types from GenericRecord fields (i.e. this is what the Avro lib will produce). The Scala collection types the ones we could drop, but are there for convenience, probably.

andreas-schroeder:
I just ran the benchmarks on master and this branch, I have bad news unfortunately:

```
with values:   Decoding.avro4sGenerated  thrpt   60  2111124,766 ± 13449,731  ops/s
Any signature: Decoding.avro4sGenerated  thrpt   60  3945842,034 ± 27771,555  ops/s
```
On simple avro records like in the benchmark, we would lose 46% throughput. 
If you want to run the benchmark on your machine, make sure that the `decode` method uses `directBinaryDecoder`, that's a mistake I made that is still on the master branch:

```scala
class Decoding extends CommonParams with BenchmarkHelpers {

  import Decoding._

  def decode[T](bytes: ByteBuffer, decoder: Decoder[T], reader: GenericDatumReader[GenericRecord]): T = {
    val dec =
      DecoderFactory.get().directBinaryDecoder(new ByteBufferInputStream(Collections.singletonList(bytes.duplicate)), null)
    val record = reader.read(null, dec)
    decoder.decode(record)
  }

  // ...
```

sksamuel:
Let me have a play with some other changes.

mberndt123:
Wow, that's pretty massive. Unfortunately it's not possible to inherit both `AnyVal` and a sealed trait… 

It should be possible to implement this as an unboxed union type:
https://medium.com/virtuslab/mechanics-of-unboxed-union-types-in-scala-647d66d34e55

The problem is that you lose some of the advantages of a sealed trait when you do this, e. g. no exhaustiveness checks in the compiler, and IntelliJ won't be able to generate the `case` clauses in a `match` expression for you.
It should be pretty easy to port this to Scala 3's union types once that is released.

sksamuel:
We can live without the exhaustiveness checks as it's a one time exercise
to make sure it's covered.

sksamuel:
Let's see how much of a difference not doing the collection conversions
gets us.

sksamuel:
How do I run the benchmarks?

andreas-schroeder:
With sbt benchmarks/jmh:run

sksamuel:
Ahh the jmh:run was what I was missing.

sksamuel:
I've played about and changed the sealed trait to a trait and made some of
the avro values be value classes. I've tried removing the extractors for
AvroRecord and the times don't budge.
Perhaps there is some other change at play that is reducing the times so
much.

sksamuel:
This latest change has performance roughly in line with the original.
I changed the pattern matching to avoid unboxing the java primitives.

andreas-schroeder:
Cool thing! Should we make the case classes final and also update the encode interface to mirror the decode interface?

andreas-schroeder:
@sksamuel @mberndt123 since we came this far now, I think we should pull this through. What do you think of my suggestion to update the encoder side?

sksamuel:
I've been doing some more testing.

It turns out, the speed of using AvroValue is dependent on the ordering of
the types in the function that converts a regular value to an instance of
the AvroValue ADT.

In other words, the match statement is compiled to a bunch of if instanceof
else checks, and so it's lightning fast when you're benchmarking with int's
and the int case is the first match. Try testing with types that are
further down the list and performance is as bad, or worse, than what
Andreas found when he first benchmarked it. And obviously not everything
can be first.

So my proposal is this. Since we are so close on 4.0.x, let's release with
what we have. The benefit of changing all the signatures now, in what is
the final release of avro4s for Scala 2.x, is probably not great.

But then we can immediately begin designing the type structure for the
Scala3/Dotty version - where anything is on the table - we can completely
change the signatures of anything - introduce functional error handling,
the AvroValue ADT and so forth. I'm aiming for a 6 month release for that.

Thoughts?

mberndt123:
:+1: sounds good to meAm 29.08.2020 02:34 schrieb Sam Sam <notifications@github.com>:

I've been doing some more testing.



It turns out, the speed of using AvroValue is dependent on the ordering of

the types in the function that converts a regular value to an instance of

the AvroValue ADT.



In other words, the match statement is compiled to a bunch of if instanceof

else checks, and so it's lightning fast when you're benchmarking with int's

and the int case is the first match. Try testing with types that are

further down the list and performance is as bad, or worse, than what

Andreas found when he first benchmarked it. And obviously not everything

can be first.



So my proposal is this. Since we are so close on 4.0.x, let's release with

what we have. The benefit of changing all the signatures now, in what is

the final release of avro4s for Scala 2.x, is probably not great.



But then we can immediately begin designing the type structure for the

Scala3/Dotty version - where anything is on the table - we can completely

change the signatures of anything - introduce functional error handling,

the AvroValue ADT and so forth. I'm aiming for a 6 month release for that.



Thoughts?







On Thu, 27 Aug 2020 at 09:50, Andreas Schroeder <notifications@github.com>

wrote:

andreas-schroeder:
Sounds like the best way to go forward!

sksamuel:
4.0.0 has been released!

PR #536, opened 2020-08-14
Fix error handling of AvroDataInputStream

### LINKED ISSUE #529: AvroDataInputStream.tryIterator can't fully consume data streams
When using the v3 api to read corrupted data using an AvroDataStream the underlying iterator from the Avro source never advances because it can't decode the record. I think this happens because in v2 the reader schema would be null if none was provided, while now a reader schema is always provided from the decoder.

Test case demonstrating non-halting:
```scala
test("reading corrupted data") {
  tryReadData[StringTest](writeData(FloatTest(3.4F)).toByteArray).foreach(println)
}
```

I wouldn't mind making an attempt at fixing this, but adding the issue in case I don't get round to making a patch.

andreas-schroeder:
Fixes #529


sksamuel:
lgtm

PR #543, opened 2020-08-28
Add support for java.time.Instant default values. Fixes #542

### LINKED ISSUE #542: Support for java.Instant default values
Using scala 2.12.6, avro4s-core 3.1.1

When defining `java.time.Instant` in top-level classes, the following exception occurs.

```scala
org.apache.avro.AvroRuntimeException: Unknown datum class: class java.time.Instant
    at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:87)
    at org.apache.avro.util.internal.JacksonUtils.toJsonNode(JacksonUtils.java:48)
    at org.apache.avro.Schema$Field.<init>(Schema.java:558)
    at com.sksamuel.avro4s.Records$.buildSchemaField(Records.scala:243)
    at com.sksamuel.avro4s.Records$.$anonfun$schema$1(Records.scala:150)
    at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:245)
    at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
    at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
    at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:38)
    at scala.collection.TraversableLike.flatMap(TraversableLike.scala:245)
```

This can be reproduced by:

```scala 
final case class ClassWithDefaultInstant(epoch: Instant = Instant.EPOCH)
```

romanskie:
Hey 

I added `java.time.Instant` top-level default value support for class definitions like:

```scala
case class ClassWithDefaultInstant(min: Instant = Instant.MIN, max: Instant = Instant.MAX, epoch: Instant = Instant.EPOCH)
```

To support `Instant.MAX` and `Instant.MIN` I used the boxed primitive max and min value from `scala.Long` in order to avoid an `java.lang.ArithmeticException: long overflow`. 

If this is an issue I need to come up with a different solution. I also thought about using a `scala.math.BigInt` or `scala.math.BigDecimal` in combination with Avro's `"logicalType": "decimal"`, but this would break the avro4s default `java.time.Instant` to `"logicalType": "timestamp-millis"` conversion. 

Please let me know if further action on my side is required :slightly_smiling_face: 

Best Regards, 
Roman

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/CustomDefaults.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/CustomDefaults.scala
@@ -1,5 +1,7 @@
 package com.sksamuel.avro4s
 
+import java.time.Instant
+
 import magnolia.{SealedTrait, Subtype}
 import org.json4s.native.JsonMethods.parse
 import org.json4s.native.Serialization.write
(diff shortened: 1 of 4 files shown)


sksamuel:
:+1:

PR #556, opened 2020-09-18
Honor custom field encoding

### LINKED ISSUE #544: Removal of writer schema on AvroOutputStream causes weird behavior
Hello guys, I have some question about a change of behavior that happened between v3.1.0 and v3.1.1.

We have the following test in our code base:

```scala
package com.arquivei.test1;

import java.io.ByteArrayOutputStream
import java.nio.ByteBuffer

import com.sksamuel.avro4s.AvroOutputStream
import org.apache.avro.Schema
import org.apache.avro.generic.{GenericDatumReader, GenericRecord}
import org.apache.avro.io.DecoderFactory
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.should.Matchers;

class Test extends AnyFlatSpec with Matchers {
  case class GenericMessage[T](Id: String, Source: String, Type: String, CreatedAt: String, Data: T)

  "test1" should "pass" in {
    val genericMessageInput: GenericMessage[Array[Byte]] = GenericMessage(
      Id = "Id",
      Source = "Source",
      Type = "Type",
      CreatedAt = "CreatedAt",
      Data = Array[Byte](0)
    )
    val genericMessageSchema =
      new Schema.Parser()
        .setValidate(false)
        .parse(
          """
            |{
            |  "type": "record",
            |  "namespace": "com.arquivei",
            |  "name": "GenericMessage",
            |  "fields" : [
            |    {"name": "Data",  "type": "bytes", "doc": "MessageContent as Avro in Wire format"},
            |    {"name": "Id", "type": "string", "doc": "Id ULID"},
            |    {"name": "CreatedAt", "type": "string", "doc": "Created Datetime of the message using RFC3339 standard"},
            |    {"name": "Source", "type": "string", "doc": "Source system that produced the message"},
            |    {"name": "Type", "type": "string", "doc": "Type of the message"}
            |  ]
            |}
            |""".stripMargin
        )

    val byteOutputStream = new ByteArrayOutputStream()
    val avroOutputStream = AvroOutputStream
      .binary[GenericMessage[Array[Byte]]]
      .to(byteOutputStream)
      .build(genericMessageSchema)
    avroOutputStream.write(genericMessageInput)
    avroOutputStream.close()
    avroOutputStream.flush()
    val encodedAvroBytes = byteOutputStream.toByteArray

    val datumReader          = new GenericDatumReader[GenericRecord](genericMessageSchema)
    val decoder              = DecoderFactory.get.binaryDecoder(encodedAvroBytes, null)
    val decodedGenericRecord = datumReader.read(null, decoder)

    decodedGenericRecord.get("Id").asInstanceOf[CharSequence].toString should be(genericMessageInput.Id)
    decodedGenericRecord.get("Source").asInstanceOf[CharSequence].toString should be(genericMessageInput.Source)
    decodedGenericRecord.get("Type").asInstanceOf[CharSequence].toString should be(genericMessageInput.Type)
    decodedGenericRecord.get("CreatedAt").asInstanceOf[CharSequence].toString should be(genericMessageInput.CreatedAt)
    decodedGenericRecord.get("Data").asInstanceOf[ByteBuffer].array should be(genericMessageInput.Data)
  }
}
```

We are basically using avro4s to encode avro data to binary, and Apache Avro libs to decode it backwards.

This test passes in version 3.1.0 but not in 3.1.1 (or later) even after doing proper changes to code. In version 3.1.1, the writer schema was removed from AvroOutputStream.build, and we removed it from the code as well. The output of this test in 3.1.1 is:

```
[info] Test:
[info] test1
[info] - should pass *** FAILED *** (206 milliseconds)
[info]   "[Source]" was not equal to "[Id]" (Test.scala:58)
[info]   Analysis:
[info]   "[Source]" -> "[Id]"
```

I don't completely understand why but it seems that avro4s was previously capable of mapping a case class to a specified writer schema using field names, but now it uses only the order declared in the case class to encode/decode data.

I understand that Avro [docs](https://avro.apache.org/docs/1.8.2/spec.html#binary_encode_complex) states that: `A record is encoded by encoding the values of its fields in the order that they are declared`, so that's a kinda expected behavior.

However, I could argue that by removing the possibility of explicitly passing a schema as argument to AvroOutputStream, the lib is forcing the user to keep two schemas in sync: a case class, and the textual schema itself. In our use case, Avro data is produced by one system and read by multiple others in several languages and we use a Schema Registry. So, it's hard to ensure that compatibility.

Am I missing something or is there any workaround for this issue?

Thanks!

andreas-schroeder:
Fixes #544

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/Records.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/Records.scala
@@ -96,8 +96,8 @@ object Records {
     val encoders = fieldsAndEncoders.map(_._2)
     update match {
       case FullSchemaUpdate(_) =>
-        record.getFields.asScala.map(f => encoders.find(e => e.fieldName == f.name).get).toVector
-        encoder.fieldEncoding = encoders.toVector
+        encoder.fieldEncoding =
+          record.getFields.asScala.map(f => encoders.find(e => e.fieldName == f.name).get).toVector
 
       case _ =>
         val fields = fieldsAndEncoders.map(_._1)
@@ -122,7 +122,6 @@ object Records {
     val decoders = fieldsAndDecoders.map(_._2)
     update match {
       case FullSchemaUpdate(_) =>
-        record.getFields.asScala.map(f => decoders.find(e => e.fieldName.contains(f.name)).get).toVector
         decoder.fieldDecoding = decoders.toVector
 
       case _ =>
@@ -260,7 +259,7 @@ object Records {
 
       val anyValSym = typeOf[AnyVal].typeSymbol
       val typeSym = sym.typeSignature.typeSymbol
-      if(typeSym.isClass && typeSym.asClass.baseClasses.contains(anyValSym)) {
+      if (typeSym.isClass && typeSym.asClass.baseClasses.contains(anyValSym)) {
         typeSym.annotations.collectFirst {
           case a if a.tree.tpe =:= typeOf[AvroDoc] =>
             val annoValue = a.tree.children.tail.head.asInstanceOf[Literal].value.value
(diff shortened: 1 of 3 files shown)


leonardoam:
Hi @andreas-schroeder , it seems that you fix satisfies solves the problem.

I would test locally if it works on my code but I couldn't get binaries for `4.1.0.1292-SNAPSHOT`. It seems your build is not working for publishing this artifacts. It's not really a problem, I can wait for your fix to be in future binaries.

sksamuel:
lgtm

PR #562, opened 2020-09-22
Add Instant logical type support

### LINKED ISSUE #549: allow java.time.Instant codec decode/encode data with respects to avro logical-type
Looks like Instant codec use fixed schema to do encode/decode [https://github.com/sksamuel/avro4s/blob/master/avro4s-core/src/main/scala/com/sksamuel/avro4s/Temporals.scala#L12](url) such that if I use .withSchema to change the logical-type of the schema to "timestamp-micros", it did not encode/decode accordingly.

One way to overcome it is to implement a customized version of Instant as mentioned in the homepage. it works but the import tax is non trivial.

is there something I overlooked?

andreas-schroeder:
Fixes #549 
Also adding logical type support for timestamp


sksamuel:
lgtm

PR #572, opened 2020-10-12
Constrain refinedSchemaFor to require a RefType

### LINKED ISSUE #571: Importing from avro4s.refined breaks certain non-refined fields/schema
Found an issue when trying to use `avro4s-refined` where it seems to affect the schema derivation for non-refined fields. I've tested with `3.0.7` and `4.0.0` and get the same with both. Examples running below are from the current master branch here.

This works as expected:
```scala
"generate correct schemas for classes with refined types" in {
  case class Test(data: Map[String, String], label: String)
  val schema = AvroSchema[Test]

  schema.getField("data").schema().getType shouldBe Schema.Type.MAP
  schema.getField("label").schema().getType shouldBe Schema.Type.STRING
}
```

However, if I add `import com.sksamuel.avro4s.refined._` the `data` field becomes a string and the test fails. This is regardless of whether the `label` field is refined or not:
* Not including `import com.sksamuel.avro4s.refined._`
```json
{
  "type" : "record",
  "name" : "Test",
  "namespace" : "unit.RefinedTest",
  "fields" : [ {
  "name" : "data",
  "type" : {
    "type" : "map",
    "values" : "string"
  }
}, {
  "name" : "label",
  "type" : "string"
} ]
}
```
* Including `import com.sksamuel.avro4s.refined._`
```json
{
  "type" : "record",
  "name" : "Test",
  "namespace" : "unit.RefinedTest",
  "fields" : [ {
  "name" : "data",
    "type" : "string"
}, {
  "name" : "label",
  "type" : "string"
} ]
}
```

If I need a refined field e.g. for `case class Test(data: Map[String, Int], label: NonEmptyString)` creating the instance manually (`implicit val labelSchemaFor: SchemaFor[NonEmptyString] = SchemaFor(Schema.create(Schema.Type.STRING))`) everything goes back to working as expected.

r-glyde:
Fixes #571 

```scala
implicit def refinedSchemaFor[T, P, F[_, _]](implicit schemaFor: SchemaFor[T]): SchemaFor[F[T, P]]
```
The above signature doesn't constrain `F[_, _]` to be a refined type. So I think when trying to resolve a schema for `Map[String, String]`, this actually matches the shape of the above instance intended for a refined type `SchemaFor[F[T, P]]` where `T` is now a string - so this just uses the `SchemaFor[String]` instance and we end up with a schema typed as a string.

By saying that `F` needs a `RefType` instance, `refinedSchemaFor` now can't be used for for a `Map[_, _]` which can now correctly resolve to `mapSchemaFor`.

### DIFF
--- a/avro4s-refined/src/main/scala/com/sksamuel/avro4s/refined/package.scala
+++ b/avro4s-refined/src/main/scala/com/sksamuel/avro4s/refined/package.scala
@@ -4,7 +4,7 @@ import eu.timepit.refined.api.{RefType, Validate}
 
 package object refined {
 
-  implicit def refinedSchemaFor[T, P, F[_, _]](implicit schemaFor: SchemaFor[T]): SchemaFor[F[T, P]] =
+  implicit def refinedSchemaFor[T, P, F[_, _] : RefType](implicit schemaFor: SchemaFor[T]): SchemaFor[F[T, P]] =
     schemaFor.forType
 
   implicit def refinedEncoder[T: Encoder, P, F[_, _] : RefType]: Encoder[F[T, P]] =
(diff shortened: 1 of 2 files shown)


sksamuel:
:+1:

PR #582, opened 2020-10-30
Refined + TypeGuardedDecoding issues

sirocchj:
Hi @sksamuel 

Been trying a few things on version 4.0.0 recently and stumbled across a set of problems when combining `Refined` and shapeless' `Coproduct`. I believe the issue is clear in the tests I've added to this.

I pinpointed the problem in the `TypeGuardedDecoding` and how - for example - `implicitly[WeakTypeTag[NonEmptyString]].tpe <:< typeOf[String]` will never be true.

The solution I propose may not be ideal but I wanted to keep the spirit of what was there already.


sksamuel:
Awesome, thanks.

PR #586, opened 2020-11-14
Allow to set default/first element of unions for sealed traits

davideicardi:
I have some records defined using a sealed trait like:

```
sealed trait Event
case object Unknown extends Event
case class Event1(...) extends Event
case class Event2(...) extends Event
```

When serialized using AVRO it will be encoded as an union of `[Event1, Event2, Unknown]`, sorting elements alphabetically.

The problem is that the first element of the union is used as the default value when the element cannot be found or is not compatible (for example if at some time I add an `Event3` class).

Using `@AvroSortPriority` I can solve this problem but it is not very "semantic". I think that, in most of the cases, only the first element is important, the other elements can be put in any order.

I propose to add a `@AvroUnionDefault` that internally can be used to force that element to be first in the list (priority=`Float.MaxValue`), but doesn't expose the internal implementation details. (maybe it can be called `@AvroTraitDefault`?)

Usage: 
```
sealed trait Event
@AvroUnionDefault
case object Unknown extends Event
case class Event1(...) extends Event
case class Event2(...) extends Event
```

I have added some tests to verify this feature and the corresponding attribute to solve it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,6 +454,43 @@ into the following AVRO `enum` schema:
 }
 ```
 
+If you just want to set the first element of the union and leave the other elements automatically 
+sorted you can use the `@AvroUnionDefault` attribute:
+
+```scala
+  sealed trait Fruit
+  @AvroUnionDefault
+  case object Unknown extends Fruit
+  case class Mango(size: Int) extends Fruit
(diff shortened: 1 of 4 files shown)


davideicardi:
Closed in favor of #587

PR #587, opened 2020-11-15
add AvroUnionPosition attributute to control position of elements in unions

davideicardi:
`AvroSortPriority` takes a `priority` to sort the elements of the union, but it is quite strange from the user perspective because the main goal is to set the elements inside the union in the order that we expect.

I have written a `@AvroUnionPosition` attribute that works in the same way but takes a `position`. It will be used like:

```scala
  sealed trait Fruit
  @AvroUnionPosition(0)
  case object Unknown extends Fruit
  @AvroUnionPosition(1)
  case class Orange(size: Int) extends Fruit
  @AvroUnionPosition(2)
  case class Mango(size: Int) extends Fruit
```

And it will produce an union with `[Unknown, Orange, Mango]`. I think it is more clear from the user perspective.

Also when we want to ensure backward/forward compatibility this can allow us to just add new elements without touching the existing ones. For example I can add an `Apple` with position `3`.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,7 +17,7 @@ jobs:
           fetch-depth: 0
 
       - name: setup sbt
-        uses: olafurpg/setup-scala@v5
+        uses: olafurpg/setup-scala@v10
 
       - name: run tests
         run: sbt ++2.12.10 test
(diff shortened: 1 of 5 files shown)


sksamuel:
Looks great. Can we add a test that if an element is not annotated it appears at the end.

davideicardi:
@sksamuel Added test for not annotated elements.
thanks!

davideicardi:
I have also updated `setup-scala` github action to v10 to fix this security issue: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Otherwise the PR workflow generate an exception, see https://github.com/sksamuel/avro4s/runs/1435858868

sksamuel:
Great work thanks.

PR #590, opened 2020-11-19
Make sys.error hygienic in macro 2.0.x

DmytroMitin:
Details are here https://stackoverflow.com/questions/64903163/implicit-resolution-fail-in-reflection-with-toolbox

Without this fix the following runtime-reflection code

```scala
toolBox.eval(q"""
    import com.sksamuel.avro4s.Decoder
    import mypackage.MyCaseClass
    Decoder.applyMacro[MyCaseClass]
  """)
```

throws

```none
scala.tools.reflect.ToolBoxError: reflective compilation has failed:
object error is not a member of package sys
```


sksamuel:
Nice catch thanks.

PR #605, opened 2020-12-22
Fix OutOfMemoryError

### LINKED ISSUE #604: java.lang.OutOfMemoryError: Java heap space
This happens when if by mistake a valid schema for some class X is used to deserialize a class A.

Below i serialize a class Test, and I use for deserialization the schema for TestInside

```
 test("Wrong schema at deserialization") {
    val avroData = new ByteArrayOutputStream()
    val os = AvroOutputStream.binary[Test].to(avroData).build()

    os.write(new Test(3, None))
    os.flush()
    os.close

    val data = avroData.toByteArray

    val testInside = com.sksamuel.avro4s.AvroSchema[TestInside].toString
    val avroTestSchema = new org.apache.avro.Schema.Parser().parse(testInside)
    val schemaFor = com.sksamuel.avro4s.SchemaFor[Test](avroTestSchema)
    val schema = com.sksamuel.avro4s.AvroSchema[Test](schemaFor)

    val outputStream = AvroInputStream.binary[Test].from(data).build(schema)
    outputStream.tryIterator.toList
    Assertions.succeed

  }

final case class TestInside(b: String)
final case class Test(a: Int, inside: Option[TestInside])
```

It is also strange that the schema is not needed for serialization.

adrian-salajan:
fixes #604
Using Try when serialization failure, the avro positions/limits are not updated so that onNext() keeps returning true.

I'm not familiar with Avro enough to know in what scenarios this tryIterator.toList could contain multiple elements like 

`List(Left(failure), Success(element))` seems kinda impossible to me

so if there is a deserialization failure the output can only be a list of one element

`List(Left(failure))`

This is a bit misleading, but at least this is a graceful failure vs OutOfMemoryError


sksamuel:
Nice job.

avro4s-core/src/main/scala/com/sksamuel/avro4s/AvroBinaryInputStream.scala
```diff
@@ -41,8 +40,15 @@ class AvroBinaryInputStream[T](in: InputStream,
     * decoding issues are wrapped.
     */
   override def tryIterator: Iterator[Try[T]] = new Iterator[Try[T]] {
-    override def hasNext: Boolean = _iter.hasNext
-    override def next(): Try[T] = Try(decoder.decode(_iter.next()))
+    var last: Option[Try[T]] = None
```
alexandru, line 43: Small issue, but you might want to use `private[this]` on members like these, because it avoids going through a setter/getter.

Next time 🙂

adrian-salajan: Thanks for the review!

PR #627, opened 2021-03-29
Added instances for Encoder/Decoder/SchemaFor for Maps that have refined strings as keys

atnoya:
Having a Map[String Refined T, A] is quite common and the only option at the moment is to convert the keys to string. This should not be required only if the type of the key is a refined String.

This PR adds support for refined keys on Maps.


sksamuel:
awesome, will do a release with this.

atnoya:
Thanks Sam!

sksamuel:
4.0.5 is syncing to maven central.

PR #637, opened 2021-04-15
Fix "namespace" in the derivation of SchemaFor for ADTs and minor fixes in examples from the documentation

pwliwanow:
Thanks for your work on avro4s!

Currently, given the following code:
```scala
@AvroNamespace("com.yuval")
sealed trait TestAnnotated
final case class TestAnnotatedImpl(value: AnnotatedNested) extends TestAnnotated

@AvroNamespace("com.yuval.nested")
case class AnnotatedNested()
```

`SchemaFor[TestAnnotated]` will produce the following schema:
```json
[ {
  "type" : "record",
  "name" : "TestAnnotatedImpl",
  "namespace" : "com.yuval",
  "fields" : [ {
    "name" : "value",
    "type" : {
      "type" : "record",
      "name" : "AnnotatedNested",
      "fields" : [ ]
    }
  } ]
} ]
```
mainly, the `namespace` field is missing from the `AnnotatedNested` record. 

Note, that `Encoder[TestAnnotated].schema` produces schema that contains `namespace` field for the `AnnotatedNested ` record:
```json
[ {
  "type" : "record",
  "name" : "TestAnnotatedImpl",
  "namespace" : "com.yuval",
  "fields" : [ {
    "name" : "value",
    "type" : {
      "type" : "record",
      "name" : "AnnotatedNested",
      "namespace" : "com.yuval.nested",
      "fields" : [ ]
    }
  } ]
} ]
```

This PR adds an appropriate test, makes `TypeUnions.schema` use the same approach as the one used for `TypeUnions.encoder` and `TypeUnions.decoder`. This PR also fixes in the documentation few examples that use `@AvroSortPriority` annotation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,7 +443,7 @@ And likewise, avro4s will convert a sealed trait such as:
 ```scala
 sealed trait Animal
 @AvroSortPriority(0) case object Cat extends Animal
-@AvroSortPriority(1) case object Dog extends Animal
+@AvroSortPriority(-1) case object Dog extends Animal
 ```
 into the following AVRO `enum` schema:
 ```json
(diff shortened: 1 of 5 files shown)


sksamuel:
Great job.

Issue #677, opened 2021-09-02
Unable to derive AvroSchema for field with list of enum as default value


KirilYurovnikRiski:
The following code:
```scala
package test

import com.sksamuel.avro4s._

sealed trait Color
case object Red extends Color
case object Green extends Color

case class Foo(colors: List[Color] = List(Red))

object Main {
  def main(args: Array[String]): Unit = {
    println(AvroSchema[Foo])
  }
}
```
fails with the following error:
```
Exception in thread "main" org.apache.avro.AvroRuntimeException: Unknown datum class: class test.Red$
	at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:96)
	at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:74)
	at org.apache.avro.util.internal.JacksonUtils.toJsonNode(JacksonUtils.java:53)
	at org.apache.avro.Schema$Field.<init>(Schema.java:586)
	at com.sksamuel.avro4s.Records$.buildSchemaField(Records.scala:242)
	at com.sksamuel.avro4s.Records$.$anonfun$schema$1(Records.scala:149)
	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:292)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:38)
	at scala.collection.TraversableLike.flatMap(TraversableLike.scala:292)
	at scala.collection.TraversableLike.flatMap$(TraversableLike.scala:289)
	at scala.collection.AbstractTraversable.flatMap(Traversable.scala:108)
	at com.sksamuel.avro4s.Records$.schema(Records.scala:143)
	at com.sksamuel.avro4s.MagnoliaDerivedSchemaFors$$anon$2.$anonfun$schemaFor$2(MagnoliaDerived.scala:30)
	at scala.Option.getOrElse(Option.scala:189)
	at com.sksamuel.avro4s.MagnoliaDerivedSchemaFors$$anon$2.schemaFor(MagnoliaDerived.scala:29)
	at com.sksamuel.avro4s.SchemaFor.resolveSchemaFor(SchemaFor.scala:53)
[...]

PR #678, opened 2021-09-04
Support array default values

kirilyuro:
Addresses https://github.com/sksamuel/avro4s/issues/677
Including various use-cases: collections and sets, optional collections and sets, unions and enums


sksamuel:
👍🏼

PR #694, opened 2021-12-31
Magnolia upgrade to V1.0.0. Fix Enums SerDe

alexlomov:
This is a draft of how new Scala 3 enums serialization/deserialization could work. This PR needs apparently a good review, given that I'm quite fresh with Avro in general and with Avro4s particularly.
Notes: 
- Scala 3 enums are translated to/from Avro enums, not unions.
- Workaround of `@AvroName`  being attached not only to the enum sealed class, but to the inheriting case classes/objects. Probably I should look into Magnolia to figure out if it is a bug or a feature.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -14,7 +14,7 @@ object Build extends AutoPlugin {
     val CatsVersion = "2.7.0"
     val RefinedVersion = "0.9.26"
     val ShapelessVersion = "2.3.7"
-    val MagnoliaVersion = "2.0.0-M9"
+    val MagnoliaVersion = "1.0.0"
     val SbtJmhVersion = "0.3.7"
     val JmhVersion = "1.32"
   }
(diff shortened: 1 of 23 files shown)


project/Build.scala
```diff
@@ -32,10 +32,11 @@ object Build extends AutoPlugin {
   override def trigger = allRequirements
   override def projectSettings = publishingSettings ++ Seq(
     organization := org,
-    scalaVersion := "3.0.2",
+    scalaVersion := "3.1.0",
```
sksamuel, line 35: do we need to build for 3.1 ?

project/Build.scala
```diff
@@ -14,7 +14,7 @@ object Build extends AutoPlugin {
     val CatsVersion = "2.7.0"
     val RefinedVersion = "0.9.26"
     val ShapelessVersion = "2.3.7"
-    val MagnoliaVersion = "2.0.0-M9"
+    val MagnoliaVersion = "1.0.0"
```
sksamuel, line 17: Did they reset versions ?

alexlomov:
Unfortunately, Magnolia 1.0.0 is built with Scala 3.1, so building Avro4s with a lower Scala version suffers of incompatibility.

alexlomov:
Yes, and provided a new groupId in.

sksamuel:
We'd need to get master working with 3.1. If so then I'm happy to merge.

alexlomov:
I can work on it a little later, like this and next week. Would that work for you?

sksamuel:
Yep that works for me

alexlomov:
Hi @sksamuel !
I've rebased my PR branch onto the most recent master. All tests pass. Is there anything else to fix?

PR #717, opened 2022-04-26
Support decoding of sealed trait + case object enums

cb372:
Added a test for this case.

Also updated the other tests in `SealedTraitDecoderTest` so they now compile, but they are all failing so I've left them commented out.

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/SealedTraitDecoderTest.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/SealedTraitDecoderTest.scala
@@ -1,125 +1,144 @@
-//package com.sksamuel.avro4s.record.decoder
-//
-//import com.sksamuel.avro4s._
-//import org.apache.avro.SchemaBuilder
-//import org.apache.avro.generic.{GenericData, GenericRecord}
-//import org.apache.avro.util.Utf8
-//import org.scalatest.funsuite.AnyFunSuite
-//import org.scalatest.matchers.should.Matchers
-//
-//class SealedTraitDecoderTest extends AnyFunSuite with Matchers {
-//
-//  test("support sealed traits of case classes") {
-//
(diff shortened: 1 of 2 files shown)


avro4s-core/src/test/scala/com/sksamuel/avro4s/record/decoder/SealedTraitDecoderTest.scala
```diff
@@ -1,125 +1,144 @@
...
+import com.sksamuel.avro4s._
+import org.apache.avro.SchemaBuilder
+import org.apache.avro.generic.{GenericData, GenericRecord}
+import org.apache.avro.util.Utf8
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class SealedTraitDecoderTest extends AnyFunSuite with Matchers {
+
+  test("support sealed trait enums") {
```
sksamuel, line 12: This passed before but not after ?
Or just commented it out by mistake ?

cb372: Sorry, my PR description was a bit terse. I was on my way out the door at the time. To summarise properly:

* The actual fix is the removal of the `&& ctx.isEnum` check in decoders/sealedtraits.scala
* This test file was completely commented out because the tests hadn't been updated to the latest API and thus didn't compile
* I uncommented the tests and updated them to make them compile, but they still failed so I commented them out again
* There was no test specifically for the 'enum encoded as sealed trait + case objects' case, so I added this one
	* There are tests like `use @AvroNamespace and @AvroName with sealed traits of case objects`, which also cover this behaviour, but those tests fail because the annotation handling hasn't been implemented yet

sksamuel: Ok thanks.

PR #720, opened 2022-05-24
Support decoding sealed trait of case classes

javimartinez:
Fix for supporting decoding sealed traits of case classes. The tests were already written, I've just made them run with success.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/decoders/unions.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/decoders/unions.scala
@@ -19,7 +19,7 @@ object TypeUnions {
       val decodersByName = ctx.subtypes.map { st =>
         val names = Names(st.typeInfo)
         val subschema = SchemaHelper.extractTraitSubschema(names.fullName, schema)
-        names.name -> st.typeclass.decode(subschema)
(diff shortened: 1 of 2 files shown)


javimartinez:
@sksamuel Could you review this? It'd be really helpful

sksamuel:
Sorry that I missed this for so long.
Will release now.

javimartinez:
Thank you!

sksamuel:
trying to release but it will only publish cats module. Don't suppose you know sbt well ?

javimartinez:
Have you fixed it? I'm not an sbt expert but if I can help with something, just let me know

sksamuel:
I just don't know it's only publishing the cats module and not the core module.

sksamuel:
There's nothing in the staging repos (it was just the cats stuff). You can
see on maven central that only the cats stuff got released for 5.0.0

coacoas:
1. Do you mean 5.0.1? I don't see 5.0.1 released anywhere, either for avro-core_3 or avro-cats_3. 
2. Do you mean staging, rather than central? I don't think I have permission to see into the staging repos.

sksamuel:
Never mind it seems to have worked. 5.0.3 is released.

PR #722, opened 2022-06-09
Update json4s to v4.0.5

alexbolo:
Updated json4s to v4.0.5 along with other dependencies: jmh, avro.
sbt 1.6.2

While upgrading our codebase to use json4s v4.x I found out that avro4s is a blocker since it's using v3.6.x.
This PR updates json4s and other dependencies to their latest versions

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -6,7 +6,7 @@ object Build extends AutoPlugin {
 
   object autoImport {
     val org = "com.sksamuel.avro4s"
-    val AvroVersion = "1.11.0"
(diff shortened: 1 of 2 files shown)


alexbolo:
@sksamuel please take a look when you have a moment

sksamuel:
BigDecimalEncoderTest failure. Maybe the encoding changed in json4s 4 ?

alexbolo:
Reverted the version of avro back to 1.9.2 and the tests are passing.
BigDecimalEncoderTest fails with avro version 1.10.2 too

Issue #741, opened 2022-10-07
broken test for LocalTime with microsecond precision


Zhen-hao:
I started doing some work for https://github.com/sksamuel/avro4s/issues/739 and saw a failure when I run the tests on `master`. 

The test `"encode and decode back to an equivalent LocalTime object when Local has microsecond precision"` in `avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue387.scala` failed with 
```
# run 1
10:48:09.356641 was not equal to 10:48:09.356641710 (GithubIssue387.scala:29)
# run 2
10:57:09.184796 was not equal to 10:57:09.184796957 (GithubIssue387.scala:29)
```

My environment:
Java: 17.0.4
sbt: 1.7.1
OS: NixOS

Also, why is the class named `GithubIssue387`?

PR #742, opened 2022-10-07
ISSUE-741 | introduce a new nanos-of-the-day logical type for LocalTime

Zhen-hao:
I found https://github.com/sksamuel/avro4s/issues/741 with Java 17.

Apparently, `LocalTime.now()` has nanosecond precision level. 
It doesn't make sense to me to cut the precision level for Avor encoding. 

We just need a new logical type for `long`.

I tested my branch with both Java 17 and Java 11.
I removed some tests because they no longer make sense because the default precision level is now nanosecond.

This change should be backward compatible because of the pattern matching on logical types.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/schemas/temporal.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/schemas/temporal.scala
@@ -14,5 +14,5 @@ trait TemporalSchemas:
   given LocalDateSchemaFor: SchemaFor[LocalDate] = DateSchemaFor.forType
   given LocalDateTimeSchemaFor : SchemaFor[LocalDateTime] = SchemaFor(TimestampNanosLogicalType.addToSchema(SchemaBuilder.builder.longType))
   given OffsetDateTimeSchemaFor : SchemaFor[OffsetDateTime] = SchemaFor(OffsetDateTimeLogicalType.addToSchema(SchemaBuilder.builder.stringType))
-  given LocalTimeSchemaFor : SchemaFor[LocalTime] = SchemaFor(LogicalTypes.timeMicros().addToSchema(SchemaBuilder.builder.longType))
+  given LocalTimeSchemaFor : SchemaFor[LocalTime] = SchemaFor(NanosOfTheDayLogicalType.addToSchema(SchemaBuilder.builder.longType))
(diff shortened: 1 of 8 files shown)


avro4s-core/src/main/scala/com/sksamuel/avro4s/schemas/temporal.scala
```diff
@@ -14,5 +14,5 @@ trait TemporalSchemas:
   given LocalDateSchemaFor: SchemaFor[LocalDate] = DateSchemaFor.forType
   given LocalDateTimeSchemaFor : SchemaFor[LocalDateTime] = SchemaFor(TimestampNanosLogicalType.addToSchema(SchemaBuilder.builder.longType))
   given OffsetDateTimeSchemaFor : SchemaFor[OffsetDateTime] = SchemaFor(OffsetDateTimeLogicalType.addToSchema(SchemaBuilder.builder.stringType))
-  given LocalTimeSchemaFor : SchemaFor[LocalTime] = SchemaFor(LogicalTypes.timeMicros().addToSchema(SchemaBuilder.builder.longType))
+  given LocalTimeSchemaFor : SchemaFor[LocalTime] = SchemaFor(NanosOfTheDayLogicalType.addToSchema(SchemaBuilder.builder.longType))
```
sksamuel, line 17: This will break/change everyone's existing schemas I think, so we need to make this opt in.

Zhen-hao: My understanding was that this change wouldn't break the decoding of data encoded by earlier versions. 

I now see one issue. A user could start producing the new schema and the new binary encoding without being aware of it. This could break downstream systems/users. 
Is that what you meant? 

I can check how the update was done last time.

Zhen-hao: @sksamuel How would you make this a user opt-in? I don't see an obvious approach.

Speaking for myself, I want this breaking change on Java 17 because the current code is already broken with it.
The change in this PR won't be an issue for me because I use schema registries everywhere and there is no way to break the decoding of data serialized with earlier versions of `avro4s`.
I also don't have the downstream issue because I can update the `avro4s` version in different systems from downstream to upstream.

Issue #768, opened 2023-03-15
No support for optional eithers


GijsbertvanVliet:
This is an issue regarding the scala3 version of avro4s, but I don't seem to be able to add the required label to this issue for some reason.
I'm trying to decode an avro event which has a field of type:
```
{
  "name": "...",
  "type":[
    "null",
    "string",
    "int"
  ]
}
```
In scala terms this would adheres to data type `Option[Either[String, Int]]`.
Unfortunately the `OptionDecoder` does not support this type since it requires the number of types in the union schema to be equal to 2. The OptionDecoder does get an `EitherDecoder` in its constructor argument though, so in theory it should be able to decode the either if it's given I think.

GijsbertvanVliet:
Ok fair enough.
But would you agree that the avro type `UNION<null,A,B>` could correspond to scala type `Option[Either[A,B]]` and similarly `UNION<null,A,B,C>` could correspond to `Option[A :+: B :+: C :+: CNil]`?
If yes, then I could try to create a PR to support these build in type mappings.

PR #772, opened 2023-03-28
Scala3/support more option types

GijsbertvanVliet:
Attempt to solve [this issue](https://github.com/sksamuel/avro4s/issues/768)
Add support for encoding and decoding of:

- option of either
- option of sealed trait of case classes
- option of sealed trait of case objects


sksamuel:
lgtm

PR #803, opened 2023-10-31
Reintroduce defaults

joan38:
This reintroduces part of the default values.
I did not reintroduce the custom types because I'm not sure if this is the right thing to do. I needs more discussion is a follow up PR IMO. But in the meantime we need this to unlock some work on our side.


sksamuel:
awesome

Issue #769, opened 2023-03-15
issue with sealed trait support


GijsbertvanVliet:
This is an issue regarding the scala3 version of avro4s, but I don't seem to be able to add the required label to this issue for some reason.
I have an issue with the support for sealed traits in avro4s. The avro schema has the following datatype:
```
{
  "type": "enum",
  "name": "NotificationTypes",
  "symbols": [
    "Unknown",
    "None",
    "Email",
    "Sms"
  ],
  "default": "Unknown"
}
```
which is converted to the following sealed trait by the avrohugger plugin:
```
sealed trait NotificationTypes

object NotificationTypes{
  case object Unknown extends NotificationTypes
  case object None    extends NotificationTypes
  case object Email   extends NotificationTypes
  case object Sms     extends NotificationTypes
}
```
The SealedTraits decoder, however, seems to scramble these values. 
It looks like this has something to do with the fact that `ctx.subtypes.sorted(SubtypeOrdering)` returns elements in order `Email, None, Sms, Unknown` while `schema.getEnumSymbols` returns elements in order `Unknown, None, Email, Sms`.

GijsbertvanVliet:
Yes, Avro hugger unfortunately does not support this. 
The problem does not occur if the `Schema` object is generated from the case classes instead of generating it from the `.avsc` file.
But even if the schema is generated from the `.avsc` file, shouldn't it be possible to match the case objects to the enum symbols purely by matching the names `Unknown`, `None`, `Email` and `Sms`?

ThijsBroersen:
I think just issue should not have been closed. The issue still exists. 
I looked into SubtypeOrdering and it looks like it is sorting one the names of the types. I think this is wrong, as the AVRO spec explicitly states the order is based on the symbols position in the enum schema: https://avro.apache.org/docs/1.11.1/specification/#sort-order

PR #804, opened 2023-11-07
fix: avro enum sorting by symbol position (as is) not alphabetical

ThijsBroersen:
This PR addresses an issue referenced in #769

Avro4s wrongfully sorts enum type symbols instead of using in the order they are defined in a schema. AVRO spec states sorting is based on symbol position: https://avro.apache.org/docs/1.11.1/specification/#sort-order

In `AvroNameSchemaTest` there was wrong testdata. `Benelux` had it's subtypes in a different order as in `avro_name_sealed_trait_symbol.json`
The test succeeded because it accidently was sorted in the correct order as `List("foofoo", "Luxembourg").sorted == List("Luxembourg", "foofoo")`, as uppercase chars precede lowercase chars.

To take into account priorities I added `EnumOrdering`, because SubtypeOrdering takes into account the type names.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,9 @@ credentials.sbt
 
 universe.avro
 sbt.json
+
+#vs-code
+.vscode
+.bloop
+.bsp
+.metals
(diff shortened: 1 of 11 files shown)


ThijsBroersen:
I think even more is broken. `@AvroSortPriority` should be used to sort in descending order. All those tests are broken now, as the priority in the test data does not match the avsc sources

PR #806, opened 2023-11-09
fix: avro enum sorting by symbol position (as is) not alphabetical

ThijsBroersen:
This PR addresses an issue referenced in #769

Avro4s wrongfully sorts enum type symbols instead of using in the order they are defined in a schema. AVRO spec states sorting is based on symbol position: https://avro.apache.org/docs/1.11.1/specification/#sort-order

In `AvroNameSchemaTest` there was wrong testdata. `Benelux` had it's subtypes in a different order as in `avro_name_sealed_trait_symbol.json`
The test succeeded because it accidently was sorted in the correct order as `List("foofoo", "Luxembourg").sorted == List("Luxembourg", "foofoo")`, as uppercase chars precede lowercase chars.

To take into account priorities I added `EnumOrdering`, because SubtypeOrdering takes into account the type names.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/decoders/sealedtraits.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/decoders/sealedtraits.scala
@@ -3,7 +3,7 @@ package com.sksamuel.avro4s.decoders
 import com.sksamuel.avro4s.{ Decoder, Avro4sConfigurationException }
 import org.apache.avro.Schema
 import com.sksamuel.avro4s.avroutils.SchemaHelper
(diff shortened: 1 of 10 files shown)


ThijsBroersen:
Magnolia 1.3.x has a reversed subtype order. Therefore I added a `.sortBy(_.index)` to guarantee enum order by symbol position

ThijsBroersen:
remark: with the new Magnolia version the java enum seems broken

sksamuel:
Nice work thanks.

ThijsBroersen:
Thanks @sksamuel , but did you notice the java enum parts? Didn't they need some work?

PR #813, opened 2023-12-18
Remove unnecessary (?) compiler dependency

dvgica:
I ended up with `protobuf-java` 3.7.0 on my project classpath, and found that the culprit was `avro4s-core -> scala3-compiler_3 -> compiler-interface`. The problem is that `protobuf-java` `3.7.0` is very old and has a number of CVEs. 

At any rate, I was confused as to why avro4s needed the `scala3-compiler_3` dependency, so tried compiling without it. It compiles fine, and tests pass.

I assume it was added for a reason, at some point, but maybe is no longer necessary? Or am I missing something?

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -38,7 +38,6 @@ object Build extends AutoPlugin {
     Test / scalacOptions ++= Seq("-Xmax-inlines:100", "-Yretain-trees"),
     javacOptions := Seq("-source", "1.8", "-target", "1.8"),    
     libraryDependencies ++= Seq(
-      "org.scala-lang"    % "scala3-compiler_3" % scalaVersion.value,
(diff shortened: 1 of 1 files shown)


sksamuel:
can't remember why it was there but I agree, doesn't seem to be needed.

dvgica:
Awesome, thanks. I wrote this PR in a bit of a rush so went back and found the [commit where it was added](https://github.com/sksamuel/avro4s/commit/c6bef313ce87f72d46b645d6808f2ffe5def5fda). Seems like it was for a `scala.tools.reflect.ToolBox`. But that code is long gone, even in the 4.x branch, and of course that compiler API doesn't exist in Scala 3. So am feeling quite confident it's safe.

PR #820, opened 2024-02-15
Remove scala-compiler dependency for release/4.0.x branch

alexklibisz:
Same as https://github.com/sksamuel/avro4s/pull/813. In my case, scala-compiler depends on a version of scala-xml that conflicts with another dependency's version of scala-xml.


sksamuel:
4.1.2 released. Thanks for PR.

PR #830, opened 2024-04-19
Match Union's sort order with enum's and with sort order in 4.x

vkorchik:
Seems like annotation `@AvroSortPriority` sorts records in Union in reversed order comparing with Enums, as well as with Enums/Unions from `avro4s:4.x`.

Probably the reason is a sort order [here in Subtypes](https://github.com/sksamuel/avro4s/blob/master/avro4s-core/src/main/scala/com/sksamuel/avro4s/typeutils/Subtypes.scala#L17). Comparing with [enums in 5.x](https://github.com/sksamuel/avro4s/blob/2beb8cbdcbb021609bd1b30f922444b81fbc0096/avro4s-core/src/main/scala/com/sksamuel/avro4s/typeutils/EnumTypes.scala#L14) and [unions in 4.x](https://github.com/sksamuel/avro4s/blob/a4aee30f04eb3f891d5d1e05e91979229dd060ac/avro4s-core/src/main/scala/com/sksamuel/avro4s/TypeUnions.scala#L103), it seems to me that the Ordering compares A with B - ascending order, when instead it should compare B to A - descending.
Since `AvroSortPriority` is descending-based (if we could say so), then there should be `B compare A`. 

The same (but reversed) issue with `AvroUnionPosition` will be automatically resolved since these two annotations are tightly coupled anyway.

### DIFF
--- a/avro4s-core/src/main/scala/com/sksamuel/avro4s/typeutils/Subtypes.scala
+++ b/avro4s-core/src/main/scala/com/sksamuel/avro4s/typeutils/Subtypes.scala
@@ -14,6 +14,6 @@ object SubtypeOrdering extends Ordering[SealedTrait.Subtype[_, _, _]] {
     val priorityA = annosA.sortPriority.getOrElse(999999F)
     val priorityB = annosB.sortPriority.getOrElse(999999F)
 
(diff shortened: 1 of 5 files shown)


vkorchik:
@sksamuel , could you please review this PR?

PR #833, opened 2024-05-14
case class schema derivation: inherit namespace from sealed trait

mberndt123:
Back in Avro4s 4, case classes inherited the `@Namespace` attribute from their parent trait. This behaviour changed in Avro4s 5, leading to all sorts of incompatible schema changes in our project.

I think the old behaviour was actually more sensible, so this PR restores it.


sksamuel:
This is a regression so I agree.

Issue #837, opened 2024-05-17
ByteBuffer.array() cannot always be used


pjfanning:
It is recommended to call ByteBuffer.hasArray before calling ByteBuffer.array()

avro4s calls array() in quite a few places.

https://github.com/search?q=repo%3Asksamuel%2Favro4s%20array()&type=code

See javadoc:
https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#array--

Calling array() on a ByteBuffer that hasArray=false (eg a read-only byte buffer) will lead to an exception.

In the cases where hasArray=false, a decision needs to be made over whether you can iterate over the bytes of the Buffer using other ByteBuffer methods or whether you could do something like wrap the ByteBuffer in an enclosing class - eg something like [ByteBufferBackedInputStream](https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-databind/2.15.2/com/fasterxml/jackson/databind/util/ByteBufferBackedInputStream.html)

sksamuel:
Makes sense. If you want to raise a PR I would cut a release.

pjfanning:
#838 was merged

PR #838, opened 2024-05-19
support ByteBuffers that have hasArray=false

pjfanning:
see #837 

I need to follow up and add some test coverage. This is just an outline of the changes.


pjfanning:
@sksamuel The tests I added do provide regression testing for buffers with hasArray=false. Do you think it is enough? I don't have a great deal of time to spend on adding extra coverage, I'm afraid.

sksamuel:
I think this looks good.

PR #839, opened 2024-05-19
standardise on UTF-8 when converting strings to/from bytes

pjfanning:
issue is that if you don't specify the charset, the default is machine specific


pjfanning:
Apologies, the code didn't compile in Java 8. I was testing locally with a newer Java version. I have added a commit that I think should fix the compile issues.

sksamuel:
lgtm

PR #884, opened 2025-11-18
OptionSchemas should not mutate original schema

### LINKED ISSUE #883: OptionSchemas mutates to the original schema
There's a bug in `OptionSchemas` - it mutates the original schema:
https://github.com/sksamuel/avro4s/blob/6f6ccc1a50ed07d602c64fc54ab52b83b301bdc9/avro4s-core/src/main/scala/com/sksamuel/avro4s/schemas/options.scala#L15-L16

Consider this example using semi-automatic derivation (`SchemaFor.derived`):

```scala
//> using scala 3.3.5
//> using dep com.sksamuel.avro4s::avro4s-core:5.0.14

import com.sksamuel.avro4s.SchemaFor

sealed trait Inner
case class A(v: String) extends Inner
case class B(v: String) extends Inner
case class Outer(v: Option[Inner])
implicit val schemaForInner: SchemaFor[Inner] = SchemaFor.derived[Inner]
println(schemaForInner.schema.getTypes.size) // 2 (union A,B)

// generating schema for Outer mutates schema for Inner:
implicit val schemaForOuter: SchemaFor[Outer] = SchemaFor.derived[Outer]
println(schemaForInner.schema.getTypes.size) // 3 (union null,A,B)

// generating it again fails as it tries to add another null
SchemaFor.derived[Outer] // throws AvroRuntimeException: Duplicate in union:null
```

pjanczyk:
Fixes #883

### DIFF
--- a/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue883.scala
+++ b/avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue883.scala
@@ -1,0 +1,20 @@
+package com.sksamuel.avro4s.github
+
+import com.sksamuel.avro4s.SchemaFor
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+class GithubIssue883 extends AnyWordSpec with Matchers {
+  "OptionSchemas" should {
+    "not mutate original schema" in {
+      sealed trait Inner
+      case class A(v: String) extends Inner
+      case class B(v: String) extends Inner
+      case class Outer(v: Option[Inner])
+      implicit val schemaForInner: SchemaFor[Inner] = SchemaFor.derived[Inner]
+      implicit val schemaForOuter: SchemaFor[Outer] = SchemaFor.derived[Outer]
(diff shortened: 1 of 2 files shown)


avro4s-core/src/test/scala/com/sksamuel/avro4s/github/GithubIssue883.scala
```diff
@@ -0,0 +1,20 @@
...
+  "OptionSchemas" should {
+    "not mutate original schema" in {
+      sealed trait Inner
+      case class A(v: String) extends Inner
+      case class B(v: String) extends Inner
+      case class Outer(v: Option[Inner])
+      implicit val schemaForInner: SchemaFor[Inner] = SchemaFor.derived[Inner]
+      implicit val schemaForOuter: SchemaFor[Outer] = SchemaFor.derived[Outer]
+      schemaForInner.schema.getTypes.size shouldBe 2 // union A,B
+      schemaForOuter.schema.getField("v").schema.getTypes.size shouldBe 3 // union null,A,B
```
pjanczyk: Before the fix, both were equal to 3

PR #904, opened 2026-04-05
Add option to use default case class value during decoding of AvroTransient field

minosiants:
Hello,
My use case is read data in protobuf format and translate it into avro format.
For protobuf I use scalapb library and it generates model case classes based on protobuf schema. 
Generated case classes have  [UnknownFieldSet](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/UnknownFieldSet) filed with default empty value. (There is an option to exclude them but I need them for analysis).
`scalapb`  allows to inject annatations into generated classes (https://scalapb.github.io/docs/customizations#adding-annotations) so to exclude `UnknownFieldSet `filed from avro schema I use `AvroTransient` annotation.  Everything works fine one model fror protobuf and avro but when `avro4s` tries to decode value that is marked with `AvroTransient` it returns None which cause class cast exception. 

I added `useDefault` boolean flag into `AvroTransient`  so now if it is true defalut value from the case class is used istead on none

### DIFF
--- a/.github/workflows/pr_scala_3.yml
+++ b/.github/workflows/pr_scala_3.yml
@@ -28,6 +28,7 @@ jobs:
         run: sbt test
 
       - name: Import GPG key
+        if: github.event.pull_request.head.repo.fork == false
         id: import_gpg
         uses: crazy-max/ghaction-import-gpg@v3
         with:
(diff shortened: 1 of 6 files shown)


minosiants:
It fails on import pgp key. Not sure if it is part of the test. I will have a look

sksamuel:
Might just need keys updated for latest central ?

sksamuel:
Merged!

